PR #36, opened 2019-08-10
Adds logging statements for two computations.

kylebgorman:
Logs the language and cutoff date strings at INFO.

### DIFF
--- a/wikipron.py
+++ b/wikipron.py
@@ -2,6 +2,7 @@
 
 import argparse
 import datetime
+import logging
 import functools
 import re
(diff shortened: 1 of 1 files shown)


wikipron.py
```diff
@@ -2,6 +2,7 @@
 
 import argparse
 import datetime
+import logging
```
jacksonllee: `logging` comes after `functools` alphabetically?

kylebgorman: Fixed

PR #37, opened 2019-08-10
Fix dialect selection

jacksonllee:
* Fixes the dialect selector. Also adds a test using US English.
* Adds a separate test for the `require_dialect_label`.

### DIFF
--- a/test_wikipron.py
+++ b/test_wikipron.py
@@ -4,6 +4,7 @@ import shutil
 import tempfile
 
 import pytest
+import requests
 import requests_html
 
 from wikipron import (
@@ -201,6 +202,16 @@ def test_li_selector(dialect, require_dialect_label, expected_li_selector):
     assert config.li_selector == expected_li_selector
 
 
+def _can_connect_to_wiktionary():
+    try:
+        requests.get("https://en.wiktionary.org/wiki/linguistics")
+    except (requests.ConnectionError, requests.ConnectTimeout):
+        return False
+    else:
+        return True
+
+
+@pytest.mark.skipif(not _can_connect_to_wiktionary(), reason="need Internet")
 def test_american_english_dialect_selection():
     # Pick a word for which Wiktionary has dialect-specified pronunciations
     # for both US and non-US English.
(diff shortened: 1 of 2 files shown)


wikipron.py
```diff
@@ -39,7 +39,7 @@
 ]
 """
 _DIALECT_SELECTOR_TEMPLATE = (
-    'span[@class = "ib-content qualifier-content"][{dialects_text}]'
+    'span[@class = "ib-content qualifier-content" and a[{dialects_text}]]'
```
jacksonllee: Here's the bug for dialect selection, where we shouldn't specified the dialect text for inside the `<a>...</a>` tags.

wikipron.py
```diff
@@ -153,7 +153,7 @@ def _get_li_selector(
             )
 
         if not dialect:
-            dialect_selector = "true"
+            dialect_selector = "true()"
```
jacksonllee: If I didn't make this change, grabbing all dialects (= not specifying any dialect, e.g., getting all pronunciations for "mocha" regardless of dialect in the test case) wouldn't work. Not sure why it didn't simply break the XPath parsing before, though.

test_wikipron.py
```diff
@@ -200,6 +201,46 @@ def test_li_selector(dialect, require_dialect_label, expected_li_selector):
...
+        **config_params, require_dialect_label=True
+    )
+    # Apply each config's XPath selector.
+    results_dialect_optional = response.html.xpath(
+        config_dialect_optional.li_selector
+    )
+    results_dialect_required = response.html.xpath(
+        config_dialect_required.li_selector
+    )
+    assert len(results_dialect_optional) > len(results_dialect_required) == 0
```
jacksonllee: `test_american_english_dialect_selection` and `test_require_dialect_label` might look pretty similar for their approach, but I've decided to keep them apart as opposed to trying to merge them as one test, for better readability.

jacksonllee: These two tests require the Internet connection, which I feel fine with. Let me know if you feel strongly about adding `pytest.mark.skipif` to avoid confusion when we run the test suite locally without Internet somehow.

kylebgorman: What's the `skipif` look like for testing for a WAN connection? I think it wouldn't be crazy---I do a lot of hobbyist development on planes and buses.

jacksonllee: The `skipif` is a decorator on top of a test that you'd like to skip if a given condition is true. It sounds you might want the connectivity check -- I've added it at c965a26. (I didn't spend too much time thinking about how to implement a WAN connectivity check, how to optimize it, etc.)

kylebgorman:
LGTM

PR #38, opened 2019-08-10
Scraping endpoint

kylebgorman:
Refactors to enable `scrape` iterator.
    
This refactors the library to expose a useful Python end-point which iterates over word/pron pairs.
    
To accomplish this, we:
    
  * eliminate argparse-specific logic from _Config so that it can be sensibly constructed from within Python
  * rename it Config so that it is no longer pseudo-private
  * refactor tests helpers and remove tests that no longer make sense
    
 This also allows us to add logging for every 100 example scraped when calling the main.


wikipron.py
```diff
@@ -225,11 +227,33 @@ def _scrape(data, config: _Config):
             if " " in pron:
                 continue
             pron = config.process_pron(pron)
-            print(f"{word}\t{pron}", file=config.output)
+            yield (word, pron)
+
+
+def scrape(config: Config) -> Iterator[Pair]:
```
jacksonllee: Given this function signature, the API design looks like this:

```python
from wikipron import Config, scrape
config = Config(...)  # kwargs needed; they are the same as those for the CLI -- run "wikipron -h" for the details
scrape(config)  # yields the (word, pronunciation) pairs
```

Should we document this in the readme?

jacksonllee: Also, any thoughts on how we can write tests for this key function?

kylebgorman: Added documentation for this usage to the README.

Not sure how to test quickly, but it's no more or less tested than before...

test_wikipron.py
```diff
@@ -26,39 +25,11 @@
...
-
-
-def test_cli_args():
-    actual_cli_args = _get_cli_args(["eng"])
-    expected_cli_args = _CLIArgs()
-    for expected_arg, expected_value in vars(expected_cli_args):
-        assert getattr(actual_cli_args, expected_arg) == expected_value
+def _config_factory(**kwargs) -> Config:
+    """Create a Config object for testing."""
+    config_dict = {"key": "eng"}  # The one default; may be overriden.
```
jacksonllee: Typo: overridden

wikipron.py
jacksonllee: Thanks for adding the docs!

Now that we have this `scrape()` function nicely decoupled from the CLI (previously it'd probably be much clunkier to test), we probably should test it, to avoid surprises down the road. I could try to come up with something in a future PR, so that we don't hold off your current one here.

kylebgorman: Agreed.

kylebgorman:
Please LGTM if you're okay with this PR.

jacksonllee:
Just checking -- is removing the `language/scrape` script intentional?

kylebgorman:
`scrape` fixed.

PR #42, opened 2019-08-12
Language codes: Handle Greek and Slovene

jacksonllee:
Here is a possible solution for handling language names like Greek and Slovene. (If this PR got in, then we'd be set for all the 21 Europarl languages.)

Main issue: Wiktionary uses [non-ISO 639 language names](https://en.wiktionary.org/wiki/Category:Terms_with_IPA_pronunciation_by_language) such as "Greek" and "Slovene" (yeah I know). This means we have two classes of solutions --

1. If we use one of the Python packages to handle ISO 639 language code to name conversion, we'll need special logic to handle the Greek/Slovene scenarios (this is what this PR demonstrates).
2. We don't use any of these ISO 639 Python packages, and we somehow curate and version our own map from ISO language codes to Wiktionary language names.

I've chosen option 1 in this PR as it's much less work. To be comprehensive for languages that we could practically get a useful amount of data for, we could look at the [Wiktionary list of languages](https://en.wiktionary.org/wiki/Category:Terms_with_IPA_pronunciation_by_language) and add to the `_LANGUAGE_CODES` dict languages with more than X entries that need similar handling. (This might sound like a lot of work, but perhaps not so, because my eyeballing so far seems to suggest that most Wiktionary language names are also ISO 639 language names and therefore no work is needed.)


jacksonllee:
@kylebgorman Any thoughts?

wikipron.py
```diff
@@ -49,6 +49,20 @@
 _PHONES_REGEX = r"\[(.+?)\]"  # FIXME: it doesn't grab anything now
 
 
+# Map from a ISO 639 code to its non-ISO 639 Wiktionary language name.
+# ISO 639-3: https://iso639-3.sil.org/sites/iso639-3/files/downloads/iso-639-3.tab  # noqa: E501
+# Wiktionary: https://en.wiktionary.org/wiki/Category:Terms_with_IPA_pronunciation_by_language  # noqa: E501
+_LANGUAGE_CODES = {
```
kylebgorman: Maybe you want to put something like:

`TODO: Expand this as needed to cover additional languages.`

kylebgorman:
I like it, modulo suggestion to mark the map as a place to insert others.

wikipron.py
jacksonllee: Added this comment at 
8f23a9e

PR #43, opened 2019-08-13
Test regexes for extracting phonetic and phonemic transcriptions

jacksonllee:
I was under the impression that `_PHONES_REGEX` didn't work and hence the `FIXME` comment in the code, but apparently I got confused and it's working just fine. This PR adds relevant tests to formally stop my nonsense.


kylebgorman:
Was wondering about that FIXME and if it was still current. Thanks for looking into this.

PR #44, opened 2019-08-13
Test the match between the CLI args and Config.__init__'s

jacksonllee:
For the CLI to work, an implicit but critical assumption is that the CLI args and their defaults are exactly those of `Config.__init__`. This PR adds a test to formally check that. (Previously I had had [this dummy class](https://github.com/kylebgorman/wikipron/pull/38/files#diff-037a040a2a7ab00ebc787ca74ebf5a53L29) to exercise the CLI interface in the test suite, which might have been overkill. I hope you think the new test in this PR is lightweight enough!)

### DIFF
--- a/wikipron.py
+++ b/wikipron.py
@@ -7,7 +7,7 @@ import logging
 import re
 import sys
 
-from typing import Callable, Iterator, Optional, TextIO, Tuple
+from typing import Callable, Iterator, List, Optional, TextIO, Tuple
 
 import iso639
 import requests
@@ -274,8 +274,7 @@ def _scrape_and_write(config: Config) -> None:
             logging.info("%d pronunciations scraped", i)
 
 
-def main() -> None:
-    logging.basicConfig(format="%(levelname)s: %(message)s", level="INFO")
+def _get_cli_args(args: List[str]) -> argparse.Namespace:
     parser = argparse.ArgumentParser(description=__doc__)
     parser.add_argument(
         "key", help="Key (i.e., name or ISO-639 code) for the language"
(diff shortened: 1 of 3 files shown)


wikipron.py
```diff
@@ -274,7 +274,7 @@ def _scrape_and_write(config: Config) -> None:
             logging.info("%d pronunciations scraped", i)
 
 
-def main() -> None:
+def _get_cli_args(args: List[str]) -> argparse.Namespace:
     logging.basicConfig(format="%(levelname)s: %(message)s", level="INFO")
```
kylebgorman: Remove this line, right?

jacksonllee: good catch -- fixed at 1ed5c00

wikipron.py
```diff
@@ -7,12 +7,15 @@
...
 
-from typing import Callable, Iterator, Optional, TextIO, Tuple
+from typing import Callable, List, Iterator, Optional, TextIO, Tuple
 
 import iso639
 import requests
 import requests_html
 
+
+__version__ = "0.1.0"
```
jacksonllee: Sneaking in this change so that `wikipron.__version__` is available as in a real released package...

jacksonllee:
@kylebgorman This PR is ready for another pass.

jacksonllee:
Never mind, I am trying to fix the commit history. Will ping again when it's really ready...

jacksonllee:
Aight, I've cleaned up the commit history here, now with two commits representing the two things I've done (test the arg match and make `__version__` available). If you LGTM this, we could do "rebase merge" so we keep master's commit history clean as well.

wikipron.py
kylebgorman: Okay with me.

PR #45, opened 2019-08-14
Fix import bug; bump version to 0.1.1

jacksonllee:
wikipron 0.1.0 is on PyPI, but isn't pip-installable. The installation hits an import error, because `setup.py` imports the `wikipron.py` module (to get `__version__` and `__doc__`), but `wikipron.py` attempts to import the third-party modules which are not yet available. This PR fixes the issue and bumps the version for a bug fix release.

@kylebgorman @elizabethgarza for viz


jacksonllee:
For documentation, 66cc416 attempted the solution of using a plain text `VERSION` file as the single source of truth for the version number, but it didn't pass the test for `$ wikipron --help` terminal command. It looks like `VERSION` wasn't available within site-packages -- probably wouldn't regardless, because wikipron is a single python file package. Reverting 66cc416.

.circleci/config.yml
```diff
@@ -5,32 +5,26 @@ jobs:
...
-            - ~/.cache/pip
       - run:
           name: Build source distribution and install package from it
+          working_directory: ~/project/
           # Ensure we can build a source distribution that can correctly install.
           # "python setup.py sdist" creates dist/wikipron-x.y.z.tar.gz
           command: |
+              pip install --progress-bar off --upgrade pip setuptools
               python setup.py sdist
               pip install dist/`ls dist/ | grep .tar.gz`
```
jacksonllee: The wikipron installation test must happen in a clean env without the third-party python packages present, or else we could run into a similar import bug again.

Removed environment cache, as building the env isn't expensive for a super light package like wikipron.

wikipron.py
```diff
@@ -9,12 +9,19 @@
...
+try:
+    import iso639
+    import requests
+    import requests_html
+except ModuleNotFoundError:
+    logging.warning(
+        "Thirty-party packages are not imported. "
+        "This situation should arise only in the installation phase, "
+        "where setup.py runs and imports this present module."
+    )
```
jacksonllee: Not super pretty, but the (an?) alternative would be to go with the usual full-blown package directory structure:

```
setup.py  # can read in _version.py without issues
wikipron/  # this directory would be in the site-packages, with _version.py inside
    __init__.py
    _version.py  # with only` __version__ = "0.1.1"`
    scrape.py
    config.py
    cli.py
tests/
    ...
```

which would seem to be a little bit too much about refactoring than fixing the import bug? I'm on the fence.

kylebgorman: Let's do it later if you want.

.circleci/config.yml
kylebgorman: SGTM.

PR #48, opened 2019-08-17
adds lexique data

m-sean:
This adds a bash script which downloads the tsv version of Lexique. It then grabs the ortho and phon transcripts (only) to match the normal data format for wikipron.


jacksonllee:
@m-sean If you're merging this PR, please use **Squash and merge** option:
Screen Shot 2019-08-19 at 11 03 54 AM
(We can't tell if github has already cached this operation for your account on this particular repo.)

PR #49, opened 2019-08-18
Handle Estonian, Slovak, etc. for title="wikipedia:<language> phonology"

jacksonllee:
For whatever reason, languages like Estonian and Slovak use `title="wikipedia:<language> phonology"` (as opposed to `Appendix:<language> pronunciation` that we've seen for other languages we've scraped so far) in the underlying HTML to signal a pronunciation. This PR expands the `<li>` XPath selector to coverage these languages. To make sure the change works, I've updated the smoke test to include Slovak.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Deprecated
 ### Removed
 ### Fixed
+- Updated the `<li>` XPath selector for `title="wikipedia:<language> phonology"`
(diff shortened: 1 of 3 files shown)


jacksonllee:
> Should we bump the version number when making a major change like this? Or should we do that in a separate PR?

A separate PR for a version number bump would allow a clean commit history, though admittedly I already did have a PR (#45...) with both code changes and a version number update. Regardless, in our present case, we still have other fixes/updates coming in before we're ready for a version bump.

kylebgorman:
SGTM

PR #50, opened 2019-08-19
Handle Korean with an extra 'span' layer under <li> XPath selector

jacksonllee:
The reason why wikipron didn't work for Korean is that Korean has an extra `<span>` layer (for specifying fonts, it looks like), which couldn't be handled by the existing XPath selector of `//li[sup[a` and would require something to the effect of `//li[span[sup[a`. This PR expands this `<li>` XPath selector to handle Korean. I wouldn't be surprised if CKJ or other languages that would need special font treatment would benefit from this change.

This PR also reactivates the `language/wikipron/scrape` script for Korean and commits the scraped data as well.


jacksonllee:
Keeping this PR as a draft, as scraping for Korean is in progress. 5500 pronunciations scraped for far; we're expecting ~12,000. 🤞  I'll also update the changelog.

PR #59, opened 2019-09-18
Skip empty pronunciations

### LINKED ISSUE #58: mssing phoneme value
Wiktionary sometimes does not provide IPAs, in which case, only graphemes are scraped.

jacksonllee:
This PR fixes the bug that we previously got scraped data where a pronunciation could be "empty" (= with only stress marks or syllable boundaries, or truly empty).

An example pronunciation in point brought up by @yeonju123 is https://en.wiktionary.org/wiki/marcap%C3%A1ginas, where the Spanish pronunciation is strangely a stress mark only as `/ˈ/` at the time of writing.

Resolves #58.

### DIFF
--- a/wikipron/config.py
+++ b/wikipron/config.py
@@ -127,16 +127,16 @@ class Config:
         if no_syllable_boundaries:
             processors.append(functools.partial(re.sub, r"\.", ""))
 
+        prosodic_markers = frozenset(["ˈ", "ˌ", "."])
+
         def wrapper(pron):
             for processor in processors:
                 pron = processor(pron)
             # GH-59: Skip prons that are empty, or have only stress marks or
             # syllable boundaries. The `any()` call is much faster than
             # re.match(r"[^ˈˌ.]", pron).
-            if any(c not in "ˈˌ." for c in pron):
+            if any(c not in prosodic_markers for c in pron):
                 return pron
-            else:
-                return None
 
         return wrapper
 
--- a/wikipron/config.py
+++ b/wikipron/config.py
@@ -130,7 +130,13 @@ class Config:
         def wrapper(pron):
             for processor in processors:
                 pron = processor(pron)
-            return pron
+            # GH-59: Skip prons that are empty, or have only stress marks or
+            # syllable boundaries. The `any()` call is much faster than
+            # re.match(r"[^ˈˌ.]", pron).
+            if any(c not in "ˈˌ." for c in pron):
+                return pron
+            else:
+                return None
 
         return wrapper
 
--- a/wikipron/scrape.py
+++ b/wikipron/scrape.py
@@ -48,10 +48,7 @@ def _scrape_once(data, config: Config) -> Iterator[Pair]:
             if " " in pron:
                 continue
             pron = config.process_pron(pron)
-            # GH-59: Skip prons that are empty, or have only stress marks or
-            # syllable boundaries. The `any()` call is much faster than
-            # re.match(r"[^ˈˌ.]", pron).
-            if any(c not in "ˈˌ." for c in pron):
+            if pron:
                 yield (word, pron)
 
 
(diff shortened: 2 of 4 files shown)


wikipron/config.py
```diff
@@ -130,7 +130,13 @@ def _get_process_pron(
         def wrapper(pron):
             for processor in processors:
                 pron = processor(pron)
-            return pron
+            # GH-59: Skip prons that are empty, or have only stress marks or
+            # syllable boundaries. The `any()` call is much faster than
+            # re.match(r"[^ˈˌ.]", pron).
```
jacksonllee: `any` is good for us here as we're only interested in "if there's any" rather than wanting some exact match. For how much faster `any` is compared to `re.match`:

```python
Python 3.7.3 (default, Mar 27 2019, 16:54:48)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.7.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import re

In [2]: %timeit re.match(r"[^ˈˌ.]", "lɪŋ.ˈɡwɪs.tɪks")
718 ns ± 1.78 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [3]: %timeit any(c not in "ˈˌ." for c in "lɪŋ.ˈɡwɪs.tɪks")
321 ns ± 1.8 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

If I used `re.match` instead of `any`, the smoke test of `wikipron.scrape` in the test suite would fail because of exceeding the timeout limit.

wikipron/scrape.py
```diff
@@ -48,7 +48,8 @@ def _scrape_once(data, config: Config) -> Iterator[Pair]:
             if " " in pron:
                 continue
             pron = config.process_pron(pron)
-            yield (word, pron)
+            if pron:
+                yield (word, pron)
```
jacksonllee: I had to implement the "pron could be `None`" logic inside `config.process_pron` for testability, or else it's hard, if not impossible, to actually test the bug fix by making `unittest.mock` work with various network API calls. The codebase as-is has the test suite outside of the library code, plus we just so happen to have a name clash between the `wikipron.scrape` and the module `scrape.py`...

wikipron/config.py
kylebgorman: what about making those three characters a `frozenset`? Even when the number of things in the set is small I've found that very very fast compared to linear comparison.

wikipron/config.py
```diff
@@ -130,7 +130,13 @@ def _get_process_pron(
...
+            if any(c not in "ˈˌ." for c in pron):
```
kylebgorman: Use a ternary here? 

   return pron if any(...) else None

Or better yet, don't explicitly return None, do something like:

    if any(...):
        return pron
    # end of function

wikipron/config.py
jacksonllee: `frozenset` does seem cleaner here. I'll switch to it. 

It looks like `frozenset` or not doesn't make a lot of difference for performance (I'm on another computer now):

```python
Python 3.6.8 |Anaconda, Inc.| (default, Dec 29 2018, 19:04:46)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.6.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: frozen = frozenset(["ˈ", "ˌ", "."])

In [2]: %timeit any(c not in "ˈˌ." for c in "lɪŋ.ˈɡwɪs.tɪks")
467 ns ± 21.4 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [3]: %timeit any(c not in frozen for c in "lɪŋ.ˈɡwɪs.tɪks")
480 ns ± 32 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [4]: %timeit any(c not in "ˈˌ." for c in "lɪŋ.ˈɡwɪs.tɪks")
457 ns ± 8.82 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [5]: %timeit any(c not in frozen for c in "lɪŋ.ˈɡwɪs.tɪks")
467 ns ± 3.52 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

wikipron/config.py
jacksonllee: I did what you see here for readability reasons, but I'm happy to simplify as well. Updating now.

Issue #57, opened 2019-09-16
[yue] Cantonese support


lfashby:
[yue] is ISO 693-3 only, but should be handled by [languagecodes.py](https://github.com/kylebgorman/wikipron/blob/master/wikipron/languagecodes.py) in the same way that [nci] for Classical Nahuatl and [hbs] for Serbo-Croatian are. 

Nonetheless, [yue] pulls nothing in phonetic and phonemic modes and does not even seem to begin scraping.

jacksonllee:
The reasons why WikiPron v0.1.1 doesn't work for Cantonese:
1. The entries under "Cantonese" on Wiktionary all use the Jyutping romanization as the head word (see [here](https://en.wiktionary.org/wiki/Category:Cantonese_terms_with_IPA_pronunciation)), which means all the entries have a numeral (which denotes tone).
2. Because WikiPron [skips entries where the head word has a numeral](https://github.com/kylebgorman/wikipron/blob/cf0e54787cf301436003436583fd80ef1cc28c82/wikipron/config.py#L185-L187), all Cantonese entries are simply skipped.

Interestingly, the entries under "Chinese" instead use the Chinese characters as the head word, and have the IPA transcriptions for various Chinese languages, including Cantonese (e.g., check out [this entry](https://en.wiktionary.org/wiki/%E9%A6%99%E6%B8%AF#Chinese) for "Hong Kong"). To scrape for Cantonese, perhaps we should target these entries under "Chinese" on Wiktionary and get the head word (= Chinese characters) as the "graphemes", and then narrow down to the Cantonese IPA pronunciation as the phonemes. (Targeting the romanization as the graphemes would be rather uninteresting for G2P...)

That being said, getting Cantonese data through WikiPron (as described in the previous paragraph) is probably not our priority right now, as the data would be less relevant for grapheme-to-phoneme modeling but more about romanization / de-romanization (is this a word?).

kylebgorman:
After talking with @lfashby I'm now convinced that the Latin and Japanese issue and the Cantonese/Chinese/Khmer/Thai issue are fundamentally the same---`Config` needs to be able to inspect `self.language` and return the appropriate selectors for both grapheme and phoneme.

jacksonllee:
#277 checked in the code to handle the scraping of Cantonese data, but we have yet to actually get the Cantonese TSV files.

kylebgorman:
I just reinstalled from HEAD and I'm getting an unrelated issue:

```
[...]

PR #61, opened 2019-10-02
TSV files for all Wiktionary languages with over 1000 entries

lfashby:
-Adds tsv files in wikipron/langauges/wikipron/tsv_files
-Adds a readme in wikpron/languages/wikipron 

The tsv file names are formatted as such: iso693-2(B)code_phonetic/phonemic
(If the language only has an iso639-3 code then that code is used. If a language doesn't have any phonetic entries on Wiktionary, then it will not have a phonetic file. Same goes for phonemic.)

The readme tsv file link links to the file (phonetic or phonemic) with more entries. I tried to determine whether or not to apply case-folding for each language, but may have gotten it wrong for a few languages. If you see any instances in the readme where I incorrectly applied or failed to apply case-folding then let me know and I can rerun those languages if need be.

I will add Russian once I have it, but it may take quite a long time to get it. I can also add all languages with more than 100 but less than 1000 entries in the same pull request as Russian if you'd like those files as well. I can submit a pull request for the code that generated all these files after that.


kylebgorman:
Cool! Could you also check in the code (I'm not seeing it on the file list but I may have missed it given how many files are involved here)? We can talk offline if there's any major concerns there or if you want to do a first-pass review before adding it to the PR.

Hey @jacksonllee (added as a reviewer) should this sort of thing go in the human-readable changelog too or should we reserve that for changes to the module itself?

kylebgorman:
One minor suggestion: `tsv_files` is redundant (what else would they be but files?) so just use `tsv`. `git mv` should be able to handle this.

lfashby:
I added the code. It is in the /src folder but it might be easier to view it by just going to my repository and looking at the tsv_files branch wikipron/languages/wikpron/src.scrape_and_write.py is what generated the tsv files and readme. codes.py and readme_insert.py both help out with certain aspects of what is needed to run scrape_and_write.py. Each has a comment at the top explaining what it does.

lfashby:
Hi there, 

I made the above changes you asked for. The new `languages/wikipron/README.md` explains broadly what the scripts do and then provides a list of steps to follow to update the dataset. 

I elected to just use `readme_insert.py` within `scrape_and_write.py` as the default method for adding to the tsv-table readme because I suppose now that the table has been made, we won't really ever be in a scenario where we only want to append to the end of it. We're either going to be updating the rows already in the table, or adding new rows throughout. 

If you have any issues with the new README please let me know. (The process for adding case-folding settings for new languages is a bit convoluted, but I'm not sure how to make it any simpler.)
I referred to the scrape function in `/wikipron/scrape.py` as the 'Wikipron scrape method' as I believe that is what we'd call such things with Javascript. Perhaps that is wrong and it would make more sense to just call it the 'Wikipron scrape function'?

kylebgorman:
Don't know anything about the Javascript world but in every language I'm familiar with a "method" is what we call class functions (which may be "instance methods" which have access to the variables of an instance of the class, "class methods" which build an instance,  or "static methods" which are defined within a class but don't get access to any instance). `def foo(x)` is a function; `class Foo: def bar(self, x)` is a method.

lfashby:
I see, so then to access the `bar` method in `class Foo: def bar(self, x)` you would use `Foo.bar(x)`.

But if I have something like:
```
import wikipron

for (word, pron) in wikipron.scrape(config):
  ...
```
Then the `.scrape(...)` is not actually method because `wikipron` isn't a class, it just happens to look like you are accessing a class function?

kylebgorman:
On Sat, Oct 5, 2019 at 2:02 PM Lucas Ashby <notifications@github.com> wrote:

> I see, so then to access the bar method in class Foo: def bar(self, x)
> you would use Foo.bar(x).
>
No, you'd construct an instance of `Foo` (call it `foo`) and then do
`foo.bar(x)`. `Foo.bar(x)` will give a syntax error (for a variety of
reasons). Can explain later.

> But if I have something like:
>
> import wikipron
>
> for (word, pron) in wikipron.scrape(config):
>   ...
>
> Then the .scrape(...) is not actually method because wikipron isn't a
> class, it just happens to look like you are accessing a class function?
>
Right, modules and classes are both namespaces as far as Python is
concerned so they have a similar syntax even if they're different things.

lfashby:
> No, you'd construct an instance of `Foo` (call it `foo`) and then do `foo.bar(x)`. `Foo.bar(x)` will give a syntax error (for a variety of reasons). Can explain later.

Right, I'm remembering a bit more about classes/objects now and that you have to 'instantiate' a class before you can use it. (I think I just used to call most things a method if you used dot notation to access it.)

I'll change the readme to say 'Wikipron's scrape function' but won't push it up unless you are all comfortable with the rest of it.

languages/wikipron/src/codes.py
```diff
@@ -0,0 +1,174 @@
...
+            isolate_language_category = re.search(
+                r"(\s+)terms(\s+)",
+                v["title"]
+            )
+            yield (
+                v["title"][0:isolate_language_category.start()],
+                num_of_pages
+            )
+        else:
+            continue
```
jacksonllee: This `else: continue` bit seems unnecessary, since the code will `continue` anyway if the if-condition is false?

languages/wikipron/src/codes.py
```diff
@@ -0,0 +1,174 @@
...
+def main():
+    new_languages = {}
+    failed_languages = {}
+    prev_languages_file = open("languages.json", "r")
+    prev_languages = json.load(prev_languages_file)
+    language_codes_file = open("iso-639-3_20190408.tsv", "r")
+    iso_list = csv.reader(language_codes_file, delimiter="\t")
+
+    for lang_page_title, total_pages in cat_members():
+        lang = {}
```
jacksonllee: It'd be nice to add a `print` line (as simple as `print(lang_page_title)`, for instance) at the beginning of the iteration, so that we see the script is chugging along.

languages/wikipron/src/codes.py
```diff
@@ -0,0 +1,174 @@
...
+            new_languages.update(lang)
+        else:
+            lang[wiktionary_code] = {
+                "wiktionary_name": wiktionary_name,
+                "total_pages": total_pages
+            }
+            failed_languages.update(lang)
+        # Reset to beginning of tsv file
+        # Else we will continue for loop from index of last match
+        language_codes_file.seek(0)
```
jacksonllee: It seems a bit odd to do this. It almost looks like we'd want to read in the language code TSV file once and keep a lookup table in memory, so as to avoid accessing the file in every iteration inside this `foo` loop. But I understand rewriting this would be a bit too substantial, probably not worth for this PR.

languages/wikipron/src/scrape_and_write.py
```diff
@@ -0,0 +1,123 @@
+import wikipron
+from readme_insert import readme_insert
+from datetime import datetime
+from time import sleep
+import requests
+import os
+import json
+
+
+def call_scrape(lang, config, file_extension):
```
jacksonllee: `file_extension` -> `filename_suffix` or similar? (The term file extension makes me think of `.tsv`, `.py`, etc. instead.)

languages/wikipron/src/scrape_and_write.py
```diff
@@ -0,0 +1,123 @@
...
+            requests.exceptions.ConnectionError
+        ) as err:
+            print(
+                "Timeout or connection error detected during scrape.",
+                lang, file_extension,
+                str(datetime.now()),
+                err
+            )
+            # Pause execution for 30 min.
+            sleep(1800)
```
jacksonllee: 30 min seems a bit long. Was there a particular reason for this choice? (I don't know how robust the Wiktionary API endpoints are.) How often did we encounter errors and hit this sleep code?

languages/wikipron/src/scrape_and_write.py
```diff
@@ -0,0 +1,123 @@
...
+            retries += 1
+            # Need to close file for it to be truncated when re-opening
+            file.close()
+            pass
```
jacksonllee: `pass` isn't needed here.

languages/wikipron/src/scrape_and_write.py
```diff
@@ -0,0 +1,123 @@
...
+        )
+
+        phonemic_count = call_scrape(iso639_code, phonemic_config, "_phonemic")
+        # Skip phonetic if phonemic failed to complete scrape
+        if phonemic_count is not None:
+            phonetic_count = call_scrape(
+                iso639_code,
+                phonetic_config,
+                "_phonetic"
+            )
```
jacksonllee: I might be missing something, but why the skip logic? If phonetic exists but phonemic is absent, I'd think we'd still want to scrape for phonetic.

languages/wikipron/src/scrape_and_write.py
```diff
@@ -0,0 +1,123 @@
...
+            print(f"TOO MANY RETRIES ON {iso639_code} MOVING ON TO NEXT.")
+            if os.path.exists(phonemic_path):
+                os.remove(phonemic_path)
+            if os.path.exists(phonetic_path):
+                os.remove(phonetic_path)
+            continue
+        # Remove files for languages that failed to call scrape altogether
+        # or for which wikipron returned nothing
+        elif phonemic_count == 0 and phonetic_count == 0:
+            print(f"FAILED TO SCRAPE {iso639_code}.")
```
jacksonllee: This was the case for Cantonese in #57, correct?

kylebgorman:
Thanks for the comments Jackson! Lucas and I went over some minor things in
person; he'll address some now and some in a later revision.

languages/wikipron/src/scrape_and_write.py
lfashby: What this does is skip calling phonetic if phonemic failed to be scraped within the set amount of retries specified in `call_scrape`. (`call_scrape` returns None in that instance.) 
 
My print messages in these if blocks are a bit confusing because there are two ways we can 'fail to scrape' a language. 
The first, we fail to complete a successful phonemic scrape within the set amount of retries (too many timeout or connection errors), `call_scrape(phonemic settings)` returns None and therefore we don't bother with a phonetic scrape.
The second, we successfully scrape with phonemic and phonetic settings, but the tsvs created are empty and `call_scrape` returns 0 for each (which is the case in your comment below regarding Cantonese).
 
I elected to skip trying a phonetic scrape if we cannot complete a phonemic scrape because it was just simpler to build something that either gets both or nothing.

languages/wikipron/src/scrape_and_write.py
lfashby: That's correct, as well as for Khmer and Thai. Both `phonetic_count` and `phonemic_count` are 0.

languages/wikipron/src/codes.py
lfashby: I agree it is quite inefficient. I'll attempt to implement a faster solution along the lines of your suggestion at some point in the future.

lfashby:
(I'm not sure if I was supposed to 'resolve' those conversations, but those that I resolved were all the suggested small changes that I addressed in this [commit](https://github.com/kylebgorman/wikipron/pull/61/commits/cffe5aee6859dc52c6bd46180cbda4b120219f5d)).

I changed all previous `print` statements to `logging.info()` statements and added logging in `codes.py` where it was suggested. An unanticipated side effect of this was that the `logging.info()` statements already within `config.py` get printed when running `scrape_and_write.py`. To avoid this, I tried changing my logging statements to `logging.debug()`and ended up printing a bunch of stuff that I think is from requests or requests_html. I decided to just use `logging.info()`, keep what is printed out in `config.py` as a way of indicating what language we are scraping, and add the name of the file in which `logging.info()` occurs to the printed out message. It ends up looking like this:

> config: 08-Oct-19 16:26:52 - Language: "Alutor"
> config: 08-Oct-19 16:26:52 - No cut-off date specified
> config: 08-Oct-19 16:27:03 - Language: "Alutor"
> config: 08-Oct-19 16:27:03 - No cut-off date specified
> scrape_and_write: 08-Oct-19 16:27:13 - 'Alutor', has no entries in phonetic transcription.

The first two config lines are generated when running with the phonemic settings and the second two in phonetic. 
If you know of any simple way to only print logging statements from a single file (so I can avoid printing the `config.py` statement and instead print something like "Currently scraping: Alutor phonemic") then let me know. Or if you think the above logging output is okay, then that works too.

I also added Russian (over 300,000 entries in a single tsv I believe) and removed Chinese.

kylebgorman:
The above logging output is okay to me. You can modify the `basicConfig`
settings to make it clearer which module or file each message comes from,
if you wish, however. See the logging docs for information about how that
all works.

jacksonllee:
Question: I see that this PR is adding 140 TSV files, but the README listing the new data shows only 72 entries but not 140. Did I miss something?

lfashby:
> Question: I see that this PR is adding 140 TSV files, but the README listing the new data shows only 72 entries but not 140. Did I miss something?

The README only links to the larger of the two tsv files obtained for each language. The reason there are not 144 tsv files is because for some languages, such as Russian (which has no phonemic entries), only one tsv file is added.

languages/wikipron/src/scrape_and_write.py
lfashby: > I elected to skip trying a phonetic scrape if we cannot complete a phonemic scrape because it was just simpler to build something that either gets both or nothing.

I'll address this and other things like adding dialect handling in version 2 while working on the suggestion to separate the scraping and writing aspects of `scrape_and_write.py`.

kylebgorman:
Maybe we shouldn't check in the smaller of the two for each language. What
do you think?

jacksonllee:
The purpose of the README is to list all the TSV files we have, and so it'd be very odd to have a TSV file but not have it listed in the README.

Re: whether we should only keep the larger one between phonemic and phonetic or keep both for a given language, I'm inclined to say we probably want both (and have them listed in the README, per the previous paragraph). IIUC, I don't think we've decided whether we'll use the phonemic or phonetic transcription for the G2P modeling we're interested in.

kylebgorman:
Actually not a bad idea to link to both. It's a value-add for WikiPron
itself.

The table is probably getting too "wide" though; maybe go for a "long"er
format with one TSV per line? Or just make sure it's readable without
scrolling on a reasonably small machine...

lfashby:
Okay, I'll add the missing row for each language and modify `scrape_and_write.py` so it looks like that is what I was intending to do all along.

lfashby:
The readme table now links to all the tsv files. 

In the future maybe I can reformat it so it is only 1 language per row or so that the second row for each language doesn't include redundant information.

kylebgorman:
I'm cool with everything at this point.

lfashby:
I don't appear to be allowed to merge it. Only those with write access can.

jacksonllee:
> I don't appear to be allowed to merge it. Only those with write access can.

Ah right, we'll take care of this for you!

PR #63, opened 2019-10-08
Use pkg_resources to get version string

### LINKED ISSUE #60: Use `pkg_resources` for version number
Use "#5" here: https://packaging.python.org/guides/single-sourcing-package-version/

jacksonllee:
Resolves #60.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,7 +41,7 @@ def main():
         packages=setuptools.find_packages(),
         python_requires=">=3.6",
         zip_safe=False,
-        install_requires=["requests", "requests-html", "iso639"],
+        install_requires=["iso639", "requests", "requests-html", "setuptools"],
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -43,7 +41,7 @@ def main():
         packages=setuptools.find_packages(),
         python_requires=">=3.6",
         zip_safe=False,
-        install_requires=["requests", "requests-html", "iso639"],
+        install_requires=["iso639", "requests", "requests-html", "setuptools"],
```
jacksonllee: We need `setuptools` in `install_requires` now strictly speaking, because `pkg_resources` is kind-of peculiar: It has its own top-level package namespace and yet isn't pip-installable itself. It is available together with `setuptools` which is a third-party package.

Issue #64, opened 2019-10-09
[scraping] Segment IPA symbols


kylebgorman:
We have found that if we construct models over Unicode codepoints, simple models will predict impossible IPA symbol combinations. (Most IPA diacritics do not exist in combining forms.) One solution @m-sean experimented with is as follows. If a codepoint is

* a Unicode combining character
* in a list of IPA "modifier" characters (e.g., ʷ)

do not split; otherwise split between codepoints. Sample logic here: https://github.com/yeonju123/pair_ngram_model/blob/master/build_sym.py#L13

I propose that, as a feature enhancement, we apply this logic in WikiPron, putting spaces between "symbols" so defined); this could also be made optional and triggered by flag.

(Then we will have to modify the baseline model tools to handle data chunked in this fashion.)

PR #69, opened 2019-10-17
add flag for whitespace delimited segementation

m-sean:
In response to scraping issue #64 (IPA symbol segmentation), I'm proposing some changes to `wikipron/cli.py` and `wikipron/config.py`. This will add a flag,`--segment`, for use on the command line. The actual execution takes place under the `_get_proccess_pron()` method in the `Config` class. This seemed like the most natural place to put it, since the pronunciations undergo other transformations there as well. I've tested the code by scraping Spanish (phonetic) and French  (phonemic) and it appears to work well.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -38,10 +38,10 @@ def test_casefold(casefold, input_word, expected_word):
 @pytest.mark.parametrize(
     "no_stress, no_syllable_boundaries, input_pron, expected_pron",
     [
-        (True, True, "lɪŋ.ˈɡwɪs.tɪks", "lɪŋɡwɪstɪks"),
-        (True, False, "lɪŋ.ˈɡwɪs.tɪks", "lɪŋ.ɡwɪs.tɪks"),
-        (False, True, "lɪŋ.ˈɡwɪs.tɪks", "lɪŋˈɡwɪstɪks"),
-        (False, False, "lɪŋ.ˈɡwɪs.tɪks", "lɪŋ.ˈɡwɪs.tɪks"),
+        (True, True, "lɪŋ.ˈɡwɪs.tɪks", "l ɪ ŋ ɡ w ɪ s t ɪ k s"),
+        (True, False, "lɪŋ.ˈɡwɪs.tɪks", "l ɪ ŋ . ɡ w ɪ s . t ɪ k s"),
+        (False, True, "lɪŋ.ˈɡwɪs.tɪks", "l ɪ ŋ ˈ ɡ w ɪ s t ɪ k s"),
+        (False, False, "lɪŋ.ˈɡwɪs.tɪks", "l ɪ ŋ . ˈ ɡ w ɪ s . t ɪ k s"),
         # GH-59: Prons with only stress or syllable boundaries are skipped.
         (False, False, "ˈ", None),
         (False, False, "", None),
@@ -53,6 +53,18 @@ def test_process_pron(
     config = config_factory(
         no_stress=no_stress, no_syllable_boundaries=no_syllable_boundaries
     )
+    assert config.process_pron(input_pron) == expected_pron
+
+
+@pytest.mark.parametrize(
+    "no_segment, input_pron, expected_pron",
+    [
+        (False, "lɛ̃.ɡɥis.tik", "l ɛ̃ . ɡ ɥ i s . t i k"),
+        (True, "lɛ̃.ɡɥis.tik", "lɛ̃.ɡɥis.tik"),
+    ],
+)
+def test_no_segment(no_segment, input_pron, expected_pron):
+    config = config_factory(no_segment=no_segment)
     assert config.process_pron(input_pron) == expected_pron
 
 
--- a/wikipron/cli.py
+++ b/wikipron/cli.py
@@ -71,6 +71,11 @@ def _get_cli_args(args: List[str]) -> argparse.Namespace:
             "overridden. If not given, results appear in stdout."
         ),
     )
+    parser.add_argument(
+        "--no-segment",
+        action="store_true",
+        help="Disable IPA segmentation with added whitespace."
+    )
     return parser.parse_args(args)
 
 
--- a/wikipron/config.py
+++ b/wikipron/config.py
@@ -7,6 +7,7 @@ import sys
 from typing import Callable, Optional, TextIO
 
 import iso639
+import unicodedata
 
 from wikipron.languagecodes import LANGUAGE_CODES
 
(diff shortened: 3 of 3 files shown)


wikipron/cli.py
```diff
@@ -71,6 +71,11 @@ def _get_cli_args(args: List[str]) -> argparse.Namespace:
             "overridden. If not given, results appear in stdout."
         ),
     )
+    parser.add_argument(
+        "--segment",
+        action="store_true",
+        help="Separate pron segments with unambiguously with added whitespace."
```
kylebgorman: Should we make this true by default? Any thoughts?

wikipron/config.py
```diff
@@ -128,6 +130,41 @@ def _get_process_pron(
...
+                "ⁿ",
+                "ʷ",
+                "˕",
+                "ˌ",
+                "˷",
+                "˔"
+            ]
+        )
+
+        def combining_mods(pron: str) -> str:
```
kylebgorman: This manipulates the combining modifiers but I'm not sure the name of the function is immediately explanatory. What about `split_combining_mod(ifiers)`?

wikipron/config.py
```diff
@@ -136,8 +173,7 @@ def wrapper(pron):
             # syllable boundaries. The `any()` call is much faster than
             # re.match(r"[^ˈˌ.]", pron).
             if any(c not in prosodic_markers for c in pron):
-                return pron
-
+                return combining_mods(pron) if segment else pron
```
kylebgorman: Shouldn't we just append `combining_mods` to the list of processors if `segment ==  True`?

m-sean: Right, that was a bit of an oversight when I added it, but now I see how simple it could be. Would it be preferable to instantiate `mods` and define the combining/modifier function directly below line 125 or would it be better to do elsewhere?

wikipron/cli.py
m-sean: I did consider it, but then it wouldn't really work to name the flag and variables `--segment` anymore, and I guess I grew attached to that. Do you have any suggestions?

wikipron/config.py
kylebgorman: I don't know where it should be defined exactly, honestly, myself. Should it even be defined in-class? Not sure. I guess somewhere before line 125, but where exactly I don't rightly know.

kylebgorman:
Fair point. Maybe we should just make it mandatory then? (This is another
possibility we discussed offline.)

wikipron/config.py
m-sean: I like the idea of keeping it partial with the processors, but the `mods` set just takes up so much space. Would it be too much to stash it somewhere and just define the combining/mods function at the top outside of the class?

kylebgorman:
I'm fine with that. Maybe document why (as a comment, not a docstring).

wikipron/config.py
jacksonllee: I think we need to test the new behavior in the test suite. I'd expect something similar to the following in `tests/test_config.py`:

```python
# I see we already have test_process_pron, but let's not overload it,
# and let's define a new test function just for this segmentation feature.

@pytest.mark.parametrize(
    "no_segment, input_pron, expected_pron",
    [
        # These are the test cases.
        (...),
        (...),
    ]
)
def test_no_segment(no_segment, input_pron, expected_pron):
    config = config_factory(no_segment=no_segment)
    assert config.process_pron(input_pron) == expected_pron
```

wikipron/config.py
```diff
@@ -128,6 +130,41 @@ def _get_process_pron(
...
+                "˔"
+            ]
+        )
+
+        def combining_mods(pron: str) -> str:
+            chars = []
+            for char in pron:
+                # Keep combining characters together
+                if unicodedata.combining(char) or char in mods:
+                    m = chars.pop()
```
jacksonllee: This `m` is the last character in `chars` that is _not_ a modifier, correct? I was confused by the name `m`. Might be nice to change its name to `last_char` or something, no strong preference though.

wikipron/config.py
```diff
@@ -128,6 +130,41 @@ def _get_process_pron(
...
+                "˳",
+                "˟",
+                "ⁿ",
+                "ʷ",
+                "˕",
+                "ˌ",
+                "˷",
+                "˔"
+            ]
+        )
```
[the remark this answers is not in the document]
kylebgorman: I think this is my fault sorta. IIRC, Sean initially used a list like that and then I suggested that he should just use the Unicode combiner codepoint feature---but it turns out many things that I as a IPA-user conceptually think of as a combiner codepoint are in fact not Unicode combiners---this is the set of such things that have the above property. (Sorry, long sentence there.) Perhaps worth adding a comment?

wikipron/config.py
```diff
@@ -54,12 +55,13 @@ def __init__(
         phonetic: bool = False,
         dialect: Optional[str] = None,
         require_dialect_label: bool = False,
+        segment: bool = True,
```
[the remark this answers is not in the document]
kylebgorman: +1

wikipron/config.py
m-sean: Good point, I will change that.

wikipron/config.py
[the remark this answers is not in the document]
m-sean: Yeah, I can see how that could be confusing. I'll add a comment about how the modifiers cover everything not in `unicode.combining`

wikipron/config.py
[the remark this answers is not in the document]
m-sean: Great! This sounds good, I'll switch this over too.

wikipron/config.py
m-sean: How about `_parse_combining_modifiers`, it seems a little more explanatory that way.

wikipron/config.py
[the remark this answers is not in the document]
m-sean: So it's still failing the tests because it defaults to segmentation, and the `expected_pron` variables are not equipped for this. 
Should I combine the tests, or should I just add spaces to the `expected_pron`'s?

wikipron/config.py
```diff
@@ -119,13 +165,16 @@ def _get_casefold(self, casefold: bool) -> Callable[[str], str]:
...
+        self, no_stress: bool, no_syllable_boundaries: bool, no_segment: bool
     ) -> Callable[[str], str]:
+
         processors = []
         if no_stress:
             processors.append(functools.partial(re.sub, r"[ˈˌ]", ""))
         if no_syllable_boundaries:
             processors.append(functools.partial(re.sub, r"\.", ""))
+        if not no_segment:
+            processors.append(functools.partial(_parse_combining_modifiers))
```
jacksonllee: The test failure is due to this line. Each "processor" is a one-argument function, and `_parse_combining_modifiers` is already one. So we don't need the `functools.partial` bit.

tests/test_config.py
```diff
@@ -56,6 +56,18 @@ def test_process_pron(
     assert config.process_pron(input_pron) == expected_pron
 
 
+@pytest.mark.parametrize(
+    "no_segment, input_pron, expected_pron",
+    [
+        (True, "lɛ̃.ɡɥis.tik", "l ɛ̃ . ɡ ɥ i . s t i k"),
+        (False, "lɛ̃.ɡɥis.tik", "lɛ̃.ɡɥis.tik"),
```
jacksonllee: Ah, the first case should be `False` for `no_segment`, and the second one `True` instead. And there's a typo in `"l ɛ̃ . ɡ ɥ i . s t i k"`: s should precede not follow the syllable boundary symbol.

Also, the test cases in `test_process_pron` need updating as well (all those cases need the spaces, since segmentation is the default behavior now).

m-sean: oops! my bad. thanks for catching.

Issue #68, opened 2019-10-16
[scraping] Support all config options


kylebgorman:
The scraping should support all config options including cut-off date, and then next big scrape run should use, IDK, 11/1/2019 as a cutoff date?

lfashby:
How should I handle the new `no_segment` option? Will we always want it set to its default value of false for all languages? If that's the case, do I need to include it at all in languages.json or can we just rely on the config.py default settings?

My inclination is to not include the `phonetic` config option, as to me part of the purpose of big scrape is to always try to scrape in both phonetic and phonemic settings for each language. 

Specifying an `output` config option prompts Wikipron to create a file that doesn't get written to (when using Wikipron from python).

jacksonllee:
For `no_segment`, I think we should leave it unspecified and let `Config` use its default settings.

+1 for getting both phonetic and phonemic data for each language. Not including `phonetic` in `languages.json` sounds good to me.

> Specifying an output config option prompts Wikipron to create a file that doesn't get written to (when using Wikipron from python).

Gah, this is a bug! I've opened #74 for this.

kylebgorman:
+1 to using default values whenever possible.

+1 to not using `phonetic` in the config.

Thanks for #74 Jackson---I was vaguely under the impression we'd taken care of that already?

lfashby:
Looking at the `_get_cut_off_date()` code in `config.py` I'm reminded I think of why I didn't use it in the original big scrape. 
If you don't specify `cut_off_date` then `_get_cut_off_date()` will generate one for you  `today = datetime.date.today()`.
If you do specify `cut_off_date` and it is later than todays date, ("2019-11-01", for example) it throws an error.
```
if d > today:
    msg = (
        "Cut-off date cannot be later than today's date: "
        f"{cut_off_date}"
    )
    raise ValueError(msg)
```
Is this the desired behavior? Wouldn't it make more sense to throw an error when `d < today`? Perhaps I am missing something.

jacksonllee:
[...]

Issue #67, opened 2019-10-16
[scraping] remove small TSVs


kylebgorman:
I propose that the big scrape, instead of removing empty files, removes any file with less than, say, 100 entries. It is relatively clear to me such files will never be any use for modeling.

I would entertain suggestions that the threshold should be 1k instead.

Issue #66, opened 2019-10-14
[eng, spa]: dialect specifications


kylebgorman:
In the current checked-in `languages.json` there is no dialect specification for either English or Spanish, whereas the earlier Bash script had one (and we want one).

For `eng`: `--dialect="US | General American"`
For `spa`:  `--dialect="Latin America"`

kylebgorman:
Once completed we want to run the scraping again for those two.

As a future step it would be good to have both British and US/GenAm English, and both Castillian/European and Latin American Spanish as separate TSVs/data sets.

lfashby:
Do you have a preference for either of these?
1. Including `require_dialect_label` for all languages (in languages.json) and setting it to false for all languages except `eng` and `spa`.
2. Omitting `require_dialect_label` for all languages except `eng` and `spa`.

(Setting `require_dialect_label` to true requires specifying a `dialect` config option as well. If we go with 1 above, then shall I also include an empty `dialect` config option for all languages?)

Neither options is better than the other when it comes to how to handle scraping for dialects in the big scrape so it really just comes down to how we want to format languages.json. 1 is perhaps better if we will be adding dialect specification to more languages in the future, 2 makes languages.json look nicer and takes advantage of config.py default settings.

jacksonllee:
Hmm, now I'm wondering if this third option may make sense: Simply remove the `require_dialect_label` option from the package altogether, i.e., if `dialect` is specified, scrape for the dialect(s) in question, otherwise just scrape for everything (whether or not the entry has a dialect label). I can see this would be a breaking change, and that we would lose the ability to scrape for all entries that explicitly have at least one dialect label, but then now I don't see why this would be useful anymore. Thoughts?

kylebgorman:
Yeah, I'm struggling to remember what the `--require_dialect_label` flag was for in the first place. Either you want to scrape a dialect (or a union of dialects), or you want to scrape everything; why would you want to scrape all labeled data but don't want to specify from which dialect? If we can't think of a good  answer, let's remove the flag.

jacksonllee:
I've just opened #75 for removing the "require dialect label" option. In general, I can help remove all obstacles for the upcoming big scrape, so that Lucas can focus on prep and config for the scrape run.

PR #76, opened 2019-10-25
Add config options to big scrape, separate scraping and writing within big scrape.

lfashby:
(Sorry for the wall of text...)
These changes are meant to address issues #66, #67, #68  as well as a few suggestions made in the comments of pull request #61.

Here are the larger changes introduced by this pull request:
- Separated the scraping and writing part of `scrape.py` (formerly `scrape_and_write.py`).
  - `write.py` now generates the README table by inspecting the contents of the `tsv/` directory. In the process it also creates a tsv `readme_tsv.tsv` with similar information as is in the README table.

- Added `no_stress`, `no_syllable_boundaries`, and `cut_off_date` options to languages in `languages.json`.
  - Modified `codes.py` to specify default values for these options when adding new languages to `languages.json` and to copy over previously set values for these options.
  - `cut_off_date` should now be set in codes.py prior to running codes.py, I’ve updated the README in languages/wikipron with those instructions.

- Added `dialect` config option (and `require_dialect_label` option) to English, Spanish and Portuguese. 
  - Restructured `scrape.py` to handle when one or more dialects are specified for a language. (Ran this new code on Portuguese, because it is a smaller language, to generate some sample data.)
  - README table now includes dialect information in Wiktionary language name column

The only small changes worth noting are:
- Logging in `scrape.py` will now also output to `scraping.log` which I’ve added to `.gitignore`. This way finding the languages that failed to be scraped is a bit easier (don’t need to scroll through the console). It also outputs the language dict from `languages.json` in the error message for languages that failed to be scraped within our set amount of retries, so it is a bit easier to build a temporary `languages.json` with the failed languages. 
- `scrape.py` will now remove files with less than 100 entries. (TSVs with less than 100 entries have been removed.)

------------------------
I have a few questions regarding dialects that I'd like your thoughts on: 

- How dialects are handled in `languages.json`.
  - I added dialects to `languages.json` in the following way: 
    ```
    "por": {
        ...
        "require_dialect_label": true,
        "dialect": {
            "_bz": "Brazil",
            "_po": "Portugal"
        },
        ...
    },
    ```
  - The keys (`_bz`, `_po`) in `dialect` serve as a sort of extension when naming the dialect tsv files (`por_bz_phonetic.tsv`, for example) and help with easy access to the dialect strings (`"Brazil"`) in `write.py`. If you'd like me to change any of the keys because you'd like different extensions for certain dialects let me know. They can be longer than two letters. I'll provide links to the English, Spanish, and Portuguese entries in `languages.json` as a separate comment so you can review the keys and dialect strings I'm using.
  - Is there a process for finding which dialects are frequently used within a given Wiktionary language category? (Aside from just checking entries and seeing whether dialect information is specified.)

- How dialects are handled in `scrape.py`.
  - As written `scrape.py` will first scrape a language entirely and then scrape for dialects if any are specified. This means it will scrape `por` (Portuguese) with no dialect, then `por` with `"Brazil"` as the dialect and then `por` with `"Portugal"` as the dialect. Is there any reason to scrape for `por` with no dialect when we are specifying a dialect? Do we want to keep the tsvs generated from previously scraping `por` (or `eng`/`spa`) with no dialect?
  - Within `scrape.py`, I moved a lot of what was in `main()` to a separate function in order to handle dialects. There may well be a better way of handling dialects than the way I've tried to do it and I'm open to suggestions on how to improve it.


lfashby:
These are what the English, Spanish and Portuguese entries look like in `languages.json`.

https://github.com/kylebgorman/wikipron/blob/faee11782214894fced4e370bc043b8792ac575a/languages/wikipron/src/languages.json#L362-L375

https://github.com/kylebgorman/wikipron/blob/faee11782214894fced4e370bc043b8792ac575a/languages/wikipron/src/languages.json#L1290-L1303

https://github.com/kylebgorman/wikipron/blob/faee11782214894fced4e370bc043b8792ac575a/languages/wikipron/src/languages.json#L1159-L1172

If you'd like, for example, the `"_ca"` in Spanish to be `"_cs"` or any other changes let me know.

lfashby:
[This is a link to the README table](https://github.com/kylebgorman/wikipron/blob/faee11782214894fced4e370bc043b8792ac575a/languages/wikipron/tsv/README.md) so you can see how Portuguese looks with dialect info in the Wiktionary name column. I don't think it would make sense to add another column for dialect info as the table is already quite wide and it would only apply to a few languages. 

[This is a link to the tsv of the README table](https://github.com/kylebgorman/wikipron/blob/faee11782214894fced4e370bc043b8792ac575a/languages/wikipron/src/readme_tsv.tsv). It contains mostly the same information as is in the README, but just lists the file name in the link column instead of a relative path in a markdown link. I could potentially add headers if that is useful. A separate column for dialects may be relevant in this format.

kylebgorman:
Just wanted to comment on some of the API choices before I dive into code. This is a lot of stuff at once, congrats on getting it all together so fast.

* I don't care for the name `readme_tsv.tsv`; it duplicates "tsv" (duh) and the fact that it is the same data as the README table is irrelevant to me. What is relevant to me as a reader is that it's a TSV file that contains the counts of data for all the languages. (`languages.tsv` maybe?)
* Can you rename `write.py` to something more informative? Maybe `summarize.py` or `generate_summary.py` since that's what it does?
* I personally have no use for the "all-dialectal" versions of English, Portuguese, and Spanish, and if someone wants them, they should just concatenate (and sort) the dialectal ones, right? So I'd suggest that if `dialect` is specified it should pre-empt an all-dialects scrape. (If someone really wanted one they could add THAT to the `dialect` specification in `languages.json`.)
* Your approach to adding dialect suffixes is fine but I would suggest that your code should add the `_` between the language and the dialect code, rather than including it in the JSON file, which feels like an abstractional leak. So it would look like 

     "dialect": { 
         "us": "US | General American", 
         "uk": "UK | Received Pronunciation" 
     }, 

but would still produce `en_us_phonemic.tsv` etc.

languages/wikipron/src/codes.py
```diff
@@ -104,12 +113,18 @@ def _scrape_wiktionary_info(lang_title):
 
 def main():
     new_languages = {}
-    failed_languages = {}
+    unmatched_languages = {}
+    basic_config_options = {
+        "casefold": None,
+        "no_stress": True,
+        "no_syllable_boundaries": True,
+        "cut_off_date": CUT_OFF_DATE
```
kylebgorman: apply reflow-er? I see there's no trailing comma here ;)

languages/wikipron/src/codes.py
```diff
@@ -104,12 +113,18 @@ def _scrape_wiktionary_info(lang_title):
...
+    }
     with open("languages.json", "r") as source:
         prev_languages = json.load(source)
     with open("iso-639-3_20190408.tsv", "r") as source:
         iso_list = csv.reader(source, delimiter="\t")
-        for (lang_page_title, total_pages) in _cat_members():
+        for (lang_page_title) in _cat_members():
```
kylebgorman: remove extraneous parens here (or I think reflow-er will take care of that)

languages/wikipron/src/codes.py
```diff
@@ -30,7 +31,13 @@
 import requests
 import requests_html
 
+# Set CUT_OFF_DATE below.
+# CUT_OFF_DATE can be no later than the date
+# at which you plan to begin scraping.
+CUT_OFF_DATE = "2019-11-01"
```
kylebgorman: I'm a little confused why this is given as a default here. Why not use `datetime.datetime.today().isoformat()` or something? (For our particular case where we really want to do a cut-off-date'd scrape we can just search-and-replace on` languages.json`, right?)

languages/wikipron/README.md
```diff
@@ -2,24 +2,26 @@
 ====================
```
kylebgorman: minor generic comment throughout: `TSV` not `tsv` (at least in comments---snake_case if it's a variable name, sure.)

languages/wikipron/src/codes.py
```diff
@@ -21,6 +21,7 @@
 * Dialect information may also need to be added manually
 """
 
+# TODO Generate and use a lookup table for the iso639-3 tsv file
```
kylebgorman: Generic minor comment throughout this change: `TSV` when you're talking about it in a comment (okay to have a variable name like `tsv` cuz snake_casing conventions).

languages/wikipron/src/scrape.py
```diff
@@ -66,99 +30,108 @@ def _call_scrape(lang, config, filename_suffix):
...
+            '"%s", returned no entries in phonemic and phonetic settings.',
+            wiki_name,
+        )
+        os.remove(phonemic_path)
+        os.remove(phonetic_path)
+        return
+    # Removes TSV files with less than 100 entries.
+    if phonemic_count < 100:
+        logging.info(
+            '"%s", has less than 100 entries in phonemic transcription.',
```
kylebgorman: what's with the comma in the log statement? This is going to say:

`Amharic, has less than 100 entries...`

Is that intended?

languages/wikipron/src/scrape.py
```diff
@@ -66,99 +30,108 @@ def _call_scrape(lang, config, filename_suffix):
...
+                    config_settings, languages[iso639_code]["wiktionary_name"],
+                    dialect_key
+                )
 
 
 if __name__ == "__main__":
     logging.basicConfig(
         format="%(filename)s %(levelname)s: %(asctime)s - %(message)s",
+        handlers=[
+            logging.FileHandler("scraping.log", mode="w"),
```
kylebgorman: possibly put filename path here into a module-level constant

languages/wikipron/src/write.py
```diff
@@ -0,0 +1,97 @@
+import os
+import json
+
+LANGUAGES_PATH = "languages.json"
+
+
+def _readme_insert(wiki_name, row_string):
+    with open("../tsv/README.md", "r+") as source:
```
kylebgorman: suggest this path as a module-level constant

languages/wikipron/src/write.py
```diff
@@ -0,0 +1,97 @@
...
+                    readme_list[i] = row_string
+                    break
+                # Inserts new row.
+                if isolated_name > wiki_name:
+                    readme_list.insert(i, row_string)
+                    break
+            # Rewrite the entire README.
+            source.seek(0)
+            source.truncate()
+            source.write("".join(readme_list))
```
kylebgorman: If each element in `readme_list` was a line w/o newlines you could use the more-efficient `writelines` (which takes an iterator of strings and doesn't build up the whole big string at once). But this part is so complex I'm afraid to ask you to kick the hornet's nest...

languages/wikipron/src/write.py
```diff
@@ -0,0 +1,97 @@
...
+        # Filters out sole README.md file
+        if file.endswith(".tsv"):
+            with open(f"{path}/{file}", "r") as tsv:
+                num_of_entries = sum(1 for line in tsv)
+            iso639_code = file[: file.index("_")]
+            transcription_level = file[
+                file.rindex("_") + 1 : file.index(".")
+            ].capitalize()
+            wiki_name = languages[iso639_code]["wiktionary_name"]
+            # Assumes we will not remove eng and spa tsv files
```
kylebgorman: Is this comment still current? Will it become irrelevant shortly? If so remove.

languages/wikipron/src/write.py
```diff
@@ -0,0 +1,97 @@
...
+            )
+            _readme_insert(wiki_name, readme_row_string)
+
+    # Write the TSV of the README
+    with open("readme_tsv.tsv", "w") as readme_tsv:
+        # Sort by wiktionary language name.
+        def sorting(ele):
+            return ele[3]
+        readme_tsv_list.sort(key=sorting)
+        for lang_row in readme_tsv_list:
```
kylebgorman: Strongly suggest you use `csv.writer` for this (you just need to tell it you're using a "\t" separator). Then call it's `writerows` method with `readme_tsv_list` as an argument and you're free of this loop too!

lfashby:
Okay, I'll start working on this tomorrow and just reply to a few comments now.

languages/wikipron/src/codes.py
lfashby: For some reason I interpreted #68 as “Language entries in languages.json should contains all config options that do not use default values set in config.py.” 

So I guess I wasted some time doing that. It would be simpler if the language entries looked like they were before: 
```
    "est": {
        "iso639_name": "Estonian",
        "wiktionary_name": "Estonian",
        "wiktionary_code": "et",
        "casefold": true
    },
```
Or this if a dialect is used:
```
    "eng": {
        "iso639_name": "English",
        "wiktionary_name": "English",
        "wiktionary_code": "en",
        "casefold": true,
        "dialect": {
            "us": "US | General American",
            "uk": "UK | Received Pronunciation"
        }
    },
```

And then `config_settings` in `scrape.py` would handle setting `no_stress`, `no_syllable_boundaries`, and `cut_off_date`, since they are always the same. Then we can just set `cut_off_date` with `datetime` in `scrape.py`.

languages/wikipron/src/write.py
lfashby: I'll look into this.

lfashby:
In addition to changing the name of `readme_tsv.tsv` I think I should also change where it gets outputted. Right now it just gets placed in the src directory, but I presume the tsv directory is the more appropriate place.

kylebgorman:
Not sure. Maybe the parent directory of src/ and tsv/ ? That way there’s no
risk of it being confused with an actual language file.

K

lfashby:
I think these changes touch on all the comments Kyle made. Aside from formatting and file name stuff, the big differences are: 

- I renamed and reworked `write.py` (now `generate_summary.py`) to be more efficient. Formally it would rewrite the README (by reading it in, splitting the rows into a list of lists, inserting a new row, converting the list of lists into a big string and writing that string) every time we iterated through a file in the tsv directory. Now it just constructs a list of lists once, converts those internal lists to strings once, and writes an all new README once. It's perhaps still not ideal, but I feel better about it.

- I added what I thought were important paths as constants to `codes.py`.

- I removed English, Spanish, and Portuguese TSVs that don't use a dialect (in preparation for the next scrape).

languages/wikipron/src/generate_summary.py
```diff
@@ -0,0 +1,76 @@
...
+            transcription_level,
+            str(num_of_entries),
+        ]
+        # TSV and README have different first column.
+        languages_summary_list.append([file] + row)
+        readme_list.append([f"[TSV]({file})"] + row)
+
+    # Sort by wiktionary language name,
+    # with phonemic entries before phonetic
+    def sorting(ele):
```
kylebgorman: still just use a lambda here.

languages/wikipron/src/scrape.py
```diff
@@ -64,101 +29,109 @@ def _call_scrape(lang, config, filename_suffix):
...
-            continue
-        # Remove files for languages that returned nothing
-        elif phonemic_count == 0 and phonetic_count == 0:
-            logging.info(
-                '"%s", returned no entries in phonemic and phonetic settings.',
-                languages[iso639_code]["wiktionary_name"],
-            )
+        if os.path.exists(phonemic_path):
             os.remove(phonemic_path)
+        if os.path.exists(phonetic_path):
```
kylebgorman: Just noticed this. This is a "dark pattern" because it is theoretically possible for the file to blink in or out of existence between the check for existence and the remove. The recommended pattern is instead to do a `try: os.remove(...)` and then catch and ignore the exception (`OSError` probably) if it fails. I realize this is an annoying one...

languages/wikipron/src/scrape.py
```diff
@@ -64,101 +29,109 @@ def _call_scrape(lang, config, filename_suffix):
...
-                "| " + " | ".join(phonetic_row) + " |\n"
-            )
-            _readme_insert(languages[iso639_code], phonetic_readme_row_string)
-        if phonemic_count:
-            phonemic_row = [f"[TSV]({iso639_code}_phonemic.tsv)"] + row
-            phonemic_row.extend(["Phonemic", str(phonemic_count)])
-            phonemic_readme_row_string = (
-                "| " + " | ".join(phonemic_row) + " |\n"
+        return
+    # Remove files for languages that returned nothing
```
kylebgorman: period for consistency on this comment I suppose

languages/wikipron/src/scrape.py
```diff
@@ -64,101 +29,109 @@ def _call_scrape(lang, config, filename_suffix):
...
+    elif phonemic_count == 0 and phonetic_count == 0:
```
kylebgorman: probably `if not phonemic_count and not phonetic_count` or something like that.

languages/wikipron/src/generate_summary.py
lfashby: As in: 
```
languages_summary_list.sort(key=lambda ele: ele[3] + ele[5])
readme_list.sort(key=lambda ele: ele[3] + ele[5])
```
Or something along the lines of: 
```
f = lambda ele: ele[3] + ele[5]
languages_summary_list.sort(key=f)
readme_list.sort(key=f)
```
The later is 'DRY-er' but isn't assigning a lambda to a variable an anti-pattern?

kylebgorman:
Former is better and hopefully the reflower can make it pretty too.

languages/wikipron/src/scrape.py
lfashby: Found this from googling:
```
from contextlib import suppress

with suppress(OSError):
    os.remove(filename)
```
Seems nicer than ignoring the exception. Let me know if you have a preference. Is it the `os.path.exists()` check that makes the file potentially blink in or out of existence? Do I need to add this to all `os.remove()` statements or just the ones that involve `os.path.exists()`

lfashby:
I'm not sure what reflower or reflow-er is and searching 'reflower python' doesn't seem to bring up anything useful.

kylebgorman:
Sorry, I’m referring to ‘white’.

lfashby:
I made the above changes and ran white. There were a few things I could do with the `os.remove` try-except block but I just went with the most straightforward method and added some comments above it.

languages/wikipron/README.md
```diff
@@ -2,17 +2,18 @@
...
 
 [scrape.py](./src/scrape.py) calls WikiPron's scraping functions on all
-Wiktionary languages with over 100 entries. It writes the results of those
-scrape calls to TSVs and generates a [README](./tsv/README.md) with selected
-information regarding the contents of those TSVs and the configuration settings
+Wiktionary languages with over 100 entries. [write.py](./src/write.py)
```
jacksonllee: `write.py` -> `generate_summary.py`?

languages/wikipron/README.md
```diff
@@ -30,9 +31,8 @@ Steps used to update the dataset
...
-    -   To find new languages running `git diff languages.json` is not
-        recommended, as it is likely that the `“total_pages”` value will have
-        updated for every language already in
-        [languages.json](./src/languages.json). Instead just search for `null`
-        values within [languages.json](./src/languages.json).
+    -   To find new languages you can run`git diff languages.json` 
+        or search for `null` values within 
+        [languages.json](./src/languages.json).
 2.  Run [scrape.py](./src/scrape.py).
+3.  Run [write.py](./src/write.py).
```
jacksonllee: `generate_summary.py`

languages/wikipron/src/generate_summary.py
```diff
@@ -0,0 +1,83 @@
...
+from codes import LANGUAGES_PATH, README_PATH, LANGUAGES_SUMMARY_PATH
+
+
+def main():
+    with open(LANGUAGES_PATH, "r") as source:
+        languages = json.load(source)
+    readme_list = []
+    languages_summary_list = []
+    path = "../tsv"
+    for file in os.listdir(path):
```
jacksonllee: `file` -> `file_path` or something?

languages/wikipron/tsv/por_bz_phonetic.tsv
```diff
@@ -0,0 +1,366 @@
...
+acontecer	a k õ t e s e ɹ
+acontecer	a k õ t e s e ɾ
+acontecer	a k õ t e s e ɻ
+acontecer	a k õ t e s e ɾ
+acontecer	a k õ t e s e χ
+actínio	a k t ʃ i n i u
+adaga	a d a ɡ a
+adenina	a d e n i n a
+adeus	a d e w s
+adição	a d͡ ʒ i s ɐ̃ w
```
jacksonllee: Oops, we forgot about having to correctly handle the tie bar for affricates. I'm opening an issue for this now.

lfashby:
> boolean coercion here?

You mean like `if not phonemic_count and not phonetic count` as in the if block below it?

I could potentially set up` _call_scrape()` so it always returns an int and then remove the files when that int is less than 100, but the idea behind having these separate if blocks and the reason `_call_scrape()` can return either an int or None is so we can log the different ways languages can fail.

kylebgorman:
I don't know if this is a good case for a returning heterogeneous types. If
you want to do that any of the following would be preferable IMO:

1. raise an exception
2. return 0 as a default
3. return -1 if it's important to distinguish between that and 0

I think #2 is probably the best option.

languages/wikipron/README.md
```diff
@@ -2,17 +2,18 @@
...
+generated and the configuration settings
 that were passed to scrape. [languages.json](./src/languages.json) provides
 [scrape.py](./src/scrape.py) with a dictionary containing the information it
 needs to call scrape on all Wiktionary languages with over 100 entries as well
 as to generate the previously mentioned [README](./tsv/README.md).
 [codes.py](./src/codes.py) is used to generate
 [languages.json](./src/languages.json). It queries Wiktionary for all languages
 with over 100 entries. It also outputs
-[failed\_langauges.json](./src/failed_languages.json), a list of languages on
+[unmatched\_langauges.json](./src/unmatched_languages.json), a list of languages on
```
kylebgorman: `langauges`

languages/wikipron/src/codes.py
```diff
@@ -104,12 +113,12 @@ def _scrape_wiktionary_info(lang_title):
 
 def main():
     new_languages = {}
-    failed_languages = {}
-    with open("languages.json", "r") as source:
+    unmatched_languages = {}
+    with open(LANGUAGES_PATH, "r") as source:
         prev_languages = json.load(source)
     with open("iso-639-3_20190408.tsv", "r") as source:
```
kylebgorman: Model level constant?

languages/wikipron/src/codes.py
```diff
@@ -122,40 +131,53 @@ def main():
...
-    with open("languages.json", "w") as json_file:
+    with open(LANGUAGES_PATH, "w") as json_file:
         json_dict = json.dumps(new_languages, indent=4)
         json_file.write(json_dict)
-    # All languages that failed to be paired with data in ISO 639 TSV file.
-    with open("failed_languages.json", "w") as failed:
-        failed_dict = json.dumps(failed_languages, indent=4)
-        failed.write(failed_dict)
+    # All languages that failed to be matched with data in ISO 639 TSV file.
+    with open("unmatched_languages.json", "w") as unmatched:
```
kylebgorman: module-level constant

languages/wikipron/src/generate_summary.py
```diff
@@ -0,0 +1,85 @@
...
+            wiki_name,
+            str(languages[iso639_code]["casefold"]),
+            transcription_level,
+            str(num_of_entries),
+        ]
+        # TSV and README have different first column.
+        languages_summary_list.append([file_path] + row)
+        readme_list.append([f"[TSV]({file_path})"] + row)
+
+    # Sort by wiktionary language name,
```
kylebgorman: `Wiktionary`

languages/wikipron/src/generate_summary.py
```diff
@@ -0,0 +1,85 @@
...
+    # with phonemic entries before phonetic.
+    languages_summary_list.sort(key=lambda ele: ele[3] + ele[5])
```
kylebgorman: Okay, so now you're using this twice, so probably define it outside of main as `_3_and_5` or something like that?

languages/wikipron/src/scrape.py
```diff
@@ -64,101 +29,92 @@ def _call_scrape(lang, config, filename_suffix):
...
-                    filename_suffix,
+                    tsv_path,
                 )
                 # Pauses execution for 10 min.
                 time.sleep(600)
+    logging.info('Failed to scrape "%s" within 10 retries. %s', lang, config)
+    return 0
+
+
+def build_config_and_filter_scrape_results(
```
kylebgorman: maybe too long, also maybe you want `_` at the beginning

kylebgorman:
Other than that, looks good to me.

PR #77, opened 2019-10-25
Remove "require_dialect_label"

### LINKED ISSUE #75: Remove the "require dialect label" option
From https://github.com/kylebgorman/wikipron/issues/66#issuecomment-544214656 and subsequent discussion, we should remove the "require dialect label" option. We would lose the ability to do the following, neither of which I think is practically useful now:
* ~One wants only entries with some dialect label, but doesn't care which dialects~ (edit: I lied -- this is not possible, because when "require dialect label" is used, "dialect" must also be specified: https://github.com/kylebgorman/wikipron/blob/abecf0341268a9426e489226599ae8881c02aaf3/wikipron/config.py#L198-L202)
* One wants a particular set of dialects (by specifying the "dialect" option) but doesn't want entries that lack a dialect label -- this seems strange, as entries without a dialect label (should) mean the pronunciation applies to all (or most major) dialects.

jacksonllee:
Resolves #75

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,13 +8,18 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## Unreleased
 
 ### Added
+- IPA segmentation using spaces by default, with the `--no-segment` flag
+  to optionally turn it off. (#69)
 - TSV files for all Wiktionary languages with over 1000 entries (except Russian) (#61)
 - Resolved Wiktionary language names for languages with at least 100
   pronunciation entries. (#52, #55)
 
 ### Changed
 ### Deprecated
+
 ### Removed
+- Removed the `require_dialect_label` option. (#77)
+
 ### Fixed
 - Skipped empty pronunciations in scraping. (#59)
 - Updated the `<li>` XPath selector for an optional layer of `<span>`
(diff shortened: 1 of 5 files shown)


jacksonllee:
@lfashby I see you have #76 on deck with "require_dialect_label" used, and hence I thought I should put this PR up to remove this option to help simplify your PR a little bit.

lfashby:
Thank you. Removing the "require_dialect_label" references in my code should be easy. Perhaps I should have given you some warning before I submitted my PR though.

jacksonllee:
No worries. Removing an option is a breaking change, and so we should probably have a separate PR like this for good software engineering practice. It's good to have your PR focus on the big scrape config work.

Tagging you for review here as well, in case you beat Kyle to it. :-)

kylebgorman:
Implementation looks good!

I had a slightly different interpretation of how things would be interpreted after the flag was removed: I thought it was going to say that if a dialect string is specified then we only sample forms that match the string (and if not specified we sample all forms). I now realize that's not what you said ('One wants a particular set of dialects (by specifying the "dialect" option) but doesn't want entries that lack a dialect label -- this seems strange, as entries without a dialect label (should) mean the pronunciation applies to all (or most major) dialects.') But I'm willing to assume that this is fine.

jacksonllee:
> I thought it was going to say that if a dialect string is specified then we only sample forms that match the string

In a way, I think this was the impetus of adding the "require_dialect_label" in the first place. :-) Since all our work is versioned, we could always revisit and possibly bring back this behavior in the future.

Issue #78, opened 2019-10-27
Handle the tie bar for affricates in IPA segmentation


jacksonllee:
IPA segmentation doesn't appear to correctly handle the tie bar for affricates. For instance, from https://github.com/kylebgorman/wikipron/pull/76#discussion_r339324885, I saw the following for Brazilian Portuguese:

```tsv
adição a d͡ ʒ i s ɐ̃ w
```

Instead, we would want the following for this example:

```tsv
adição a d͡ʒ i s ɐ̃ w
```

m-sean:
this seems like a job for the parser i wrote! happy to submit a pull request for it to resolve the issue.

PR #79, opened 2019-10-28
handle tiebar in IPA segmentation 

m-sean:
I've written a fix for issue #78 accompanied with a test in the segmentation portion of `test_config.py`. Apologies for the oversight when I originally added this in!


wikipron/config.py
```diff
@@ -73,6 +73,9 @@ def _parse_combining_modifiers(pron: str) -> str:
         if unicodedata.combining(char) or char in UNICODE_MODIFIERS:
             last_char = chars.pop()
             chars.append(f"{last_char}{char}")
+        elif chars and re.match(r"\w[͜͡]$", chars[-1]):
```
kylebgorman: Could you add a one-line comment here to explain how this works?

PR #87, opened 2019-11-06
Uses `segments` tokenization.

### LINKED ISSUE #83: Potential problem in _parse_combining_modifiers()
I started the second big scrape and while scraping for phonetic data from Albanian, Wikipron threw an error, the last line of which I'll reproduce below:

>   File ".../wikipron/config.py", line 73, in _parse_combining_modifiers
>     last_char = chars.pop()
> IndexError: pop from empty list

The final line in the Albanian phonetic tsv is `herë	h ɛː ɾ` meaning the scrape likely failed [on this entry](https://en.wiktionary.org/wiki/hidrogjen#Albanian) which contains what looks like word initial aspiration. 

I guess for words like the one that caused this error we would want to combine with next char `ʰi d r ɔ ɟ ɛ n`?

kylebgorman:
To resolve issues of how to tokenize IPA strings, we defer to
`segments.Tokenizer`. In the future we can build custom "profiles" or
bug the authors to fix bugs.

Closes #83.


setup.py
```diff
@@ -41,7 +41,13 @@ def main():
         packages=setuptools.find_packages(),
         python_requires=">=3.6",
         zip_safe=False,
-        install_requires=["iso639", "requests", "requests-html", "setuptools"],
+        install_requires=[
+            "iso639",
+            "segments",
```
jacksonllee: It looks like they follow semantic versioning (see [their changelog](https://github.com/cldf/segments/blob/393aef48423ed56660052ea583d2e13bd3fcd3e4/CHANGES.md)), and development is still pretty active (last commit on master was from yesterday). Perhaps we should have `"segments < 3"` here, just in case they have API-breaking changes?

(I think the other packages in `install_requires` are pretty stable and therefore it's fine to not have an upper bound for version numbers.)

kylebgorman: Didn't realize `install_requires` supported that. Done.

kylebgorman:
Added the recommended tests and the version info to setup. PTAL.

kylebgorman:
Done.

PR #88, opened 2019-11-06
Generalized word and pron extraction; handle Khmer

### LINKED ISSUE #70: [khm] Khmer support
Khmer IPA transcriptions are within tables. [Example](https://en.wiktionary.org/wiki/កងទ័ពជើងទឹក#Khmer)

An obvious, although perhaps not ideal, solution would simply be to create a new selector template in `config.py` that targets `<tr>` or `<td>` elements containing a `span[@class = "IPA"]`.  You'd then have to have some logic in the `yield_phn` function that checks if the language is Khmer or Thai and switches the selector template.

jacksonllee:
This PR refactors the scraping logic for a generalized (or at least generalizable) word and pron extraction. As a test case, Khmer (whose pronunciations are embedded in HTML tables instead of in `<li>` bullets for most other languages on Wiktionary) is resolved with its own language-specific pronunciation extraction.

Resolves #70


wikipron/extract/__init__.py
```diff
@@ -0,0 +1,10 @@
+from wikipron.extract.default import extract_word_pron_default
+from wikipron.extract.khm import extract_word_pron_khmer
+
+
+# All extraction functions must have the exact same function signature.
+EXTRACTION_FUNCTIONS = {
+    "default": extract_word_pron_default,
+    # Key has to be the language name used by Wiktionary.
+    "Khmer": extract_word_pron_khmer,
+}
```
jacksonllee: Here's the high-level idea: For every language we need special support for, define its own "extract_word_pron_X" function and add it here.

wikipron/config.py
```diff
@@ -135,36 +138,24 @@ def wrapper(pron):
...
-                return None
-            # Skips examples containing digits.
-            if re.search(r"\d", word):
-                return None
-            # Skips examples available later than the cut-off date.
-            if date > cut_off_date:
-                return None
-            return word
-
-        return wrapper
```
jacksonllee: This bit is now separated between the `_skip_word` and `_skip_date` functions, which I now strongly prefer to my previous convoluted version here that tied word and date together.

wikipron/scrape.py
```diff
@@ -15,41 +15,35 @@
...
-            except IndexError:
-                continue
-            # Removes parens around various segments.
-            pron = pron.replace("(", "").replace(")", "")
-            # Skips examples with a space in the pron.
-            if " " in pron:
-                continue
-            pron = config.process_pron(pron)
-            if pron:
-                yield (word, pron)
```
jacksonllee: This bit is moved to a core `yield_pron` function.

wikipron/extract/default.py
```diff
@@ -0,0 +1,19 @@
...
+
+def _yield_phn(request, config):
+    for pron_xpath in request.html.xpath(config.pron_xpath_selector):
+        yield from yield_pron(pron_xpath, IPA_XPATH, config)
+
+
+def extract_word_pron_default(word, request, config):
+    words = itertools.repeat(config.casefold(word))
+    prons = _yield_phn(request, config)
+    yield from zip(words, prons)
```
jacksonllee: With only Khmer as a non-default case, I'm not entirely sure if this is the right level of abstraction for various things. For this PR, I'm happy with the way it is. (I do not plan to handle other special languages within this PR.) When we add more languages under `wikipron/extract/`, we may have some refactoring here.

tests/test_scrape.py
```diff
@@ -1,32 +1,93 @@
...
+
+Parameters
+----------
+key : str
+    An ISO 639 code or language name.
+wik_name : str
+    The language name used by Wiktionary.
+config_params : dict
+    Parameters for the Config class.
+"""
```
jacksonllee: I'm used to the [numpydoc format](https://numpydoc.readthedocs.io/en/latest/format.html) for docstrings simply because I'm in the data science world now with extensive use of the pydata stack (sklearn, etc.) all documented in this style. If you guys feel strongly about using another docstring protocol, I'd be happy to comply. (IMO, numpydoc is clearer than the (default?) reStructuredText format.)

kylebgorman: I am fine with this entirely, so long as we standardize. (And I think we should add more typing too.)

wikipron/config.py
kylebgorman: Nice, this is a lot clearer, yes.

wikipron/extract/__init__.py
kylebgorman: +1

wikipron/extract/khm.py
```diff
@@ -0,0 +1,14 @@
+"""Word and pron extraction for Khmer."""
+
+import itertools
+
+from wikipron.extract.default import yield_pron
+
+
+_IPA_XPATH = '//span[@class = "IPA" and @lang = "km"]'
```
kylebgorman: This is pseudo-private because it's just for Khmer, is that right?

Could the khm module JUST contain the `IPA_XPATH` spec? OR could we just have `wikipron/extract/ipa_xpath.py` which containts `IPA_XPATH_DEFAULT`, `IPA_XPATH_KHMER`, and then momve `extract_word_pron` to `core.py`?

wikipron/extract/default.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+IPA_XPATH = '//span[@class = "IPA"]'
+
+
+def _yield_phn(
+    request: requests.Response, config: "Config"
+) -> "Iterator[Pron]":
+    for pron_xpath in request.html.xpath(config.pron_xpath_selector):
+        yield from yield_pron(pron_xpath, IPA_XPATH, config)
```
jacksonllee: Comparing the default treatment and Khmer, one common denominator is what I've abstracted out as the `yield_pron` function (defined in `wikipron/extract/core.py`. The default case needs an extra "layer" of looping over what we previously called the "li_selector".

`yield_pron` should hopefully be generic enough and usable by other exceptional languages.

Why isn't there a `yield_word`? Both default and Khmer simply use the head word from a Wiktionary entry page, and therefore no special treatment is needed. Latin would likely need its own `yield_word`, for grabbing the desired graphemes with macrons.

jacksonllee:
@kylebgorman Thanks for the LGTM! I've made a couple updates based on your comments. Let me know if you want further changes. If not, I'll go ahead and merge this in (or you could do it, too).

PR #89, opened 2019-11-08
Tightens versioning for `segments`.

kylebgorman:
Upstream, `segments` incorporated [a fix we proposed](https://github.com/cldf/segments/commit/d74a23c53bd9f25f5a90f45a5068e6284d4a7708). I make sure the version we use is at least that high. I then revert [a minor change made to avoid the underlying issue](https://github.com/cldf/segments/commit/d74a23c53bd9f25f5a90f45a5068e6284d4a7708).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,5 +4,5 @@ pytest==5.0.1
 pytest-timeout==1.3.3
 requests==2.21.0
 requests-html==0.9.0
-segments==2.1.1
+segments>=2.1.2,<3
 setuptools==41.4.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,5 +4,5 @@ pytest==5.0.1
 pytest-timeout==1.3.3
 requests==2.21.0
 requests-html==0.9.0
-segments>=2.1.2,<3
+segments==2.1.2
 setuptools==41.4.0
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -4,5 +4,5 @@ pytest==5.0.1
 pytest-timeout==1.3.3
 requests==2.21.0
 requests-html==0.9.0
-segments==2.1.1
+segments>=2.1.2,<3
```
jacksonllee: This should be `segments==2.1.2`.

Projects vary for how `requirements.txt` is used (between whether it's an app or a library? WikiPron is a library). For WikiPron, it's been the source of truth for setting up a deterministic environment (so with exact version pins) for both [local development](https://github.com/kylebgorman/wikipron/blame/260ba9d99e9b99197d4ec5b402031b5a6ff742cf/README.md#L105) and [CircleCI builds](https://github.com/kylebgorman/wikipron/blob/260ba9d99e9b99197d4ec5b402031b5a6ff742cf/.circleci/config.yml#L29) (but not in "production", which could be anyone's personal computer or whatever for a library), while `setup.py`'s `install_requires` state the minimal requirements (so with version ranges or no version restrictions -- `segments>=2.1.2,<3` in `setup.py` below is good). FWIW, we're following the [official guidelines](https://packaging.python.org/discussions/install-requires-vs-requirements/#requirements-files) for differentiating `requirements.txt` and `setup.py`'s `install_requires`.

kylebgorman:
Done.

PR #90, opened 2019-11-08
Added Thai pron extraction

lfashby:
Added `tha.py` to extract directory. It is almost identical to `khm.py`, the only difference is in the `_IPA_XPATH` variable, which oddly enough, can be set to the same value as `IPA_XPATH` in `default.py`. The difference between `tha.py` and `default.py` is that we don't want to use `_yield_phn` for `tha.py` as we want to skip the `for pron_xpath in request.html.xpath(config.pron_xpath_selector):` step in `_yield_phn` and just go straight to calling the core `yield_pron` function.
I suppose I could import `IPA_XPATH` from `default.py` into `tha.py` if that would make more sense?

I'm checking in the Thai data as well, it's our first data collected using the new 'segments' package parsing. 

I also changed logging in scrape.py to just log messages from within scrape.py (as well as a few changes I noticed should be made running the big scrape).


jacksonllee:
Sorry I forgot this -- please add an entry to CHANGELOG.md for handling Thai.

lfashby:
+1 for having a general `--no-tone` ticket. 

The changelog entry is basically the same as Khmer's, I couldn't think of anything more creative.

PR #92, opened 2019-11-08
Handle Latin

### LINKED ISSUE #80: [lat] Macrons not scraped
The Latin "big scrape" files are missing macrons (vertical bars above vowels indicating contrastive length) on the grapheme side. This is because, in Wiktionary backend, Latin words WITHOUT macrons are used as headwords. Example: https://en.wiktionary.org/wiki/malus#Etymology_2_2

Note that that page does pair the macroned form _mālus_ with thhe pronunciation `maːlus` like we want it to. It seems to me that we need to switch from using the backend headword to the word rendered on the page.

(As it happens, this is exactly the same issue that we discuss for the CoNLL-SIGMORPHON 2017 shared task in http://wellformedness.com/papers/gorman-etal-2019.pdf, which also scraped data from Wiktionary.)

jacksonllee:
This PR resolves #80 for Latin.

I tested Latin scraping with `$ wikipron lat > lat.tsv` and the result TSV file has 25,731 entries (compared to 32,526, the number of entries for which Wiktionary says Latin should have pronunciation).

To illustrate what the TSV file would look like with the "malus" example from Kyle, the code correctly targeted the three "Etymologies" and their orthographies (with and without macrons) plus the corresponding pronunciation.

```
malus   ˈm a . l u s
mālus   ˈm aː . l u s
mālus   ˈm aː . l u s
```


tests/test_scrape.py
```diff
@@ -33,12 +33,13 @@
...
     SmokeTestLanguage("khm", "Khmer", {}),
     SmokeTestLanguage("tha", "Thai", {}),
+    SmokeTestLanguage("lat", "Latin", {}),
 ]
 
 
 @pytest.mark.skipif(not can_connect_to_wiktionary(), reason="need Internet")
 @pytest.mark.parametrize("smoke_test_language", _SMOKE_TEST_LANGUAGES)
-@pytest.mark.timeout(3)
+@pytest.mark.timeout(4)
```
jacksonllee: I could get the Latin smoke test to pass the smoke test under 3 seconds for timeout. I'm writing up a ticket with some thoughts about performance in general.

wikipron/extract/lat.py
```diff
@@ -0,0 +1,101 @@
...
+1. The default extraction function uses the Wiktionary entry page title
+   as the graphemes. Latin uses the macrons orthographically (for vowel
+   length), but the Wiktionary entry page titles never have them.
+2. Relatedly, because the orthographic distinction by macrons is collapsed,
+   a Latin entry page organizes the "homographs" in terms of "Etymologies".
+   Each etymology has its own (correct) orthographic form and pronunciations.
+
+The solution for Latin is to go through each "Etymology" and extract the
+word and pronunciation at this level.
+"""
```
jacksonllee: Is the solution (both code and documentation) readable? While it works, I'm also concerned if it'd be too cryptic for our future selves.

wikipron/extract/lat.py
```diff
@@ -0,0 +1,108 @@
...
+    # We will get the word from each "Etymology" section within the page.
+    word = None  # noqa: F841
+    etymology_tags = _get_etymology_tags(request)
+    for etymology_tag in etymology_tags:
+        # The words and prons are extracted from the same request response but
+        # separately (so with somewhat overlapping XPath selectors), because
+        # the targeted words and prons are at the same hierarchical in the
+        # underlying HTML, and may be separated by other irrelevant sibling
+        # tags. Trying to get both words and prons while walking through
+        # the request response only once might be technically be, but the
```
lfashby: might be technically possible?

wikipron/extract/lat.py
```diff
@@ -0,0 +1,108 @@
...
+) -> "Iterator[WordPronPair]":
+    # For Latin, we don't use the title word from the Wiktionary page,
+    # because it never has macrons (necessary for Latin vowel length).
+    # We will get the word from each "Etymology" section within the page.
+    word = None  # noqa: F841
+    etymology_tags = _get_etymology_tags(request)
+    for etymology_tag in etymology_tags:
+        # The words and prons are extracted from the same request response but
+        # separately (so with somewhat overlapping XPath selectors), because
+        # the targeted words and prons are at the same hierarchical in the
```
lfashby: at the same hierarchical level? 
(Sorry about these nitpicks.)

jacksonllee: This is why I need code reviews :) Please do keep comments coming!

jacksonllee:
> do you want to wait for Kyle to look it over as well?

Yeah -- I'm expecting Kyle (a long time researcher on Latin phonology and morphology) to be ecstatic about the Latin data getting fixed. ;)

wikipron/extract/__init__.py
```diff
@@ -1,6 +1,7 @@
 from wikipron.extract.default import extract_word_pron_default
 from wikipron.extract.khm import extract_word_pron_khmer
 from wikipron.extract.tha import extract_word_pron_thai
+from wikipron.extract.lat import extract_word_pron_latin
```
kylebgorman: I know this is petty but, can you lexicographically sort this?

wikipron/extract/__init__.py
```diff
@@ -9,4 +10,5 @@
     # Key has to be the language name used by Wiktionary.
     "Khmer": extract_word_pron_khmer,
     "Thai": extract_word_pron_thai,
+    "Latin": extract_word_pron_latin,
```
kylebgorman: Ditto

PR #93, opened 2019-11-10
More Latin

### LINKED ISSUE #80: [lat] Macrons not scraped

jacksonllee:
#92 handled Latin for entries with explicit "Etymology" sections. Apparently, there are lots of entries without such sections ([example](https://en.wiktionary.org/wiki/abace#Latin)), though these latter entries still have the problem that we'd have to extract the correct word form from within the page. This PR recovers these missed entries.

#92 could get us ~26K Latin entries, but we'd have ~36K entries from this PR (scraping tested locally).

(This PR more thoroughly resolves #80.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ Unreleased
 ### Added
 
 -   Handled Latin, for which the actual graphemes cannot be the Wiktionary
-    page titles and have to come from within the page. (\# 92)
+    page titles and have to come from within the page. (\# 92, \# 93)
 -   Handled Thai, whose pronunciations are embedded in HTML tables. (\#90)
 -   Handled Khmer, whose pronunciations are embedded in HTML tables. (\#88)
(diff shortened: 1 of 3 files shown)


jacksonllee:
My local scraping test didn't account for the potential duplicated entries that Lucas's de-duping post-processing script would take care of. We shall see...

Issue #84, opened 2019-11-01
[kxd] Brunei (Malay) not in languagecodes.json


lfashby:
Trying to scrape for Brunei (ISO 639 name; Brunei Malay is the Wiktionary name) led to the following error. 

>   File ".../wikipron/config.py", line 128, in _get_language
>     language = iso639.to_name(key).split(";")[0]
>   File ".../lib/python3.7/site-packages/iso639/__init__.py", line 115, in to_name
>     raise NonExistentLanguageError('Language does not exist.')
> iso639.NonExistentLanguageError: Language does not exist.

Brunei Malay's code is "kxd" which is ISO 639-3 only, meaning it isn't covered by the iso639 package we are using. 

For the time being I'll just go ahead and add `"kxd": "Brunei Malay"` myself to languagecodes.py. Perhaps it should be added in a separate pull request from the one I'll eventually submit with all the new data?
 I'm not sure how this happened, shouldn't this have been caught by test_languagecodes.py? (Unless someone added 228 entries to Brunei Malay since test_languagecodes.py was last run.)

jacksonllee:
> I'm not sure how this happened, shouldn't this have been caught by test_languagecodes.py? (Unless someone added 228 entries to Brunei Malay since test_languagecodes.py was last run.)

FWIW, the master branch (updated 2 days ago) is now associated with [this build for the Python 3.6 run](https://circleci.com/gh/kylebgorman/wikipron/198?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link), with the following warnings:

```
tests/test_languagecodes.py::test_language_coverage
  /root/project/tests/test_languagecodes.py:97: UserWarning: WikiPron cannot handle "Brunei Malay".
    warnings.warn(f'WikiPron cannot handle "{language}".')

tests/test_languagecodes.py::test_language_coverage
  /root/project/tests/test_languagecodes.py:97: UserWarning: WikiPron cannot handle "Mecayapan Nahuatl".
    warnings.warn(f'WikiPron cannot handle "{language}".')
```

The `languagecodes.py` module was added maybe ~2 months ago, without Brunei Malay or Mecayapan Nahuatl. So it's probably true that the community has been adding pronunciation entries to Wiktionary from time to time, and that some languages which didn't pass the threshold of >= 100 pronunciation entries before now do and trigger the warning in the build.

[...]

PR #95, opened 2019-11-14
Data for all languages with over 100 entries, except phonetic French and Russian.

lfashby:
I also reverted the changes I made to logging in `scrape.py` in my previous PR and updated `languagecodes.py` in accordance with #84 .

Latin, Khmer, Thai and 3 small languages (Choctaw, Tetelcingo Nahuatl, Mecayapan Nahuat) were run with the new `segments` parsing. 

Does this PR require a changelog entry? We do have data for new languages (previous big scrape only included languages with over 1000 entries) and the new data for languages that we previously scrapped is different in that it has been parsed and the duplicates are removed.

I'll add Russian and French when I get them, but wanted to submit this now to make working on Serbo-Croatian easier.


lfashby:
I'm not sure how these changes could cause the tests to timeout. Is this related to the performance problems that Jackson commented on [in the Latin PR](https://github.com/kylebgorman/wikipron/pull/92#discussion_r344353474)?

kylebgorman:
FYI it looks like a few of the smoke tests just timed out (which happens).

Issue #94, opened 2019-11-12
Big scrape directory reorg


kylebgorman:
Taking this online after discussion with @lfashby...

One annoying thing about the [big scrape scripts](https://github.com/kylebgorman/wikipron/tree/master/languages/wikipron) is that the TSV directory has a huge list of files only then followed by the README table below. One way to work around this is as follows:

* `wikipron/languages/wikipron/README.md`, which details how to re-run the big scrape, should be moved to the `src` subdirectory, which doesn't have a README yet
* `wikipron/languages/wikipron/tsv/README.md`, the autogenerated language table, should be moved to `wikipron/languages/wikipron` instead.

Since this will screw with the paths we should wait until after the next submission deadline to do it...

PR #96, opened 2019-11-17
Serbo-Croatian data splitting and readme reorganization.

### LINKED ISSUE #62: [hbs] Serbo-Croatian support
Wiktionary only lists Serbo-Croatian and does not have separate categories for Serbian, Croatian, Montenegrin, and Bosnian - the languages that the ISO 693-3 [hbs] code encapsulates. 

Wikipron can handle scraping if given [hbs] but the resulting tsv files may not be of much use. [hbs] tsv files are therefore not included in pull request #61 .

lfashby:
Resolves #62 and #94 

The solution for splitting the Serbo-Croatian (hbs) data is perhaps a bit crude, it merely checks each grapheme for Latin characters and if it finds any, writes the grapheme and pron to the Latin tsv. Seems to do the trick. I wrote it so that it handles both phonetic and phonemic files, even though it doesn't appear there are any (or at least aren't enough) Serbo-Croatian phonetic entries on Wikitionary thus far.

I moved both README's and reworked the links, and also wrote some instructions in the big scrape README for running the big scrape on a subset of languages.


languages/wikipron/languages_summary.tsv
```diff
@@ -174,7 +174,8 @@ sco_phonemic.tsv	sco	Scots	Scots	True	Phonemic	869
 sco_phonetic.tsv	sco	Scots	Scots	True	Phonetic	420
 gla_phonemic.tsv	gla	Gaelic; Scottish Gaelic	Scottish Gaelic	True	Phonemic	904
 gla_phonetic.tsv	gla	Gaelic; Scottish Gaelic	Scottish Gaelic	True	Phonetic	118
-hbs_phonemic.tsv	hbs	Serbo-Croatian	Serbo-Croatian	True	Phonemic	46370
+hbs_cyrillic_phonemic.tsv	hbs	Serbo-Croatian	Serbo-Croatian (Cyrillic)	True	Phonemic	22690
+hbs_latin_phonemic.tsv	hbs	Serbo-Croatian	Serbo-Croatian (Latin)	True	Phonemic	23680
```
kylebgorman: I realize this seems silly but `cyrl` and `latn` are standard abbreviations for this.

languages/wikipron/src/README.md
```diff
@@ -0,0 +1,77 @@
...
+    -   If there are new Wiktionary languages with over 100 entries, they will
+        be added to [languages.json](languages.json).
+    -   As mentioned in the comment at the top of [codes.py](codes.py),
+        any new languages added to [languages.json](languages.json) will
+        not have their case-folding settings specified. 
+        Their `"casefold"` value will therefore be set to `None`.
+    -   Whether or not to apply case-folding for these new languages needs
+        to be manually set by changing the "casefold" value within
+        [languages.json](languages.json).
+    -   To find new languages you can run`git diff languages.json`
```
kylebgorman: Missing space

languages/wikipron/src/README.md
```diff
@@ -0,0 +1,77 @@
...
+        to be manually set by changing the "casefold" value within
+        [languages.json](languages.json).
+    -   To find new languages you can run`git diff languages.json` 
+        or search for `null` values within 
+        [languages.json](languages.json).
+2.  Run [scrape.py](scrape.py).
+    -   By default `cut_off_date` in `main()` is set using
+        `datetime.date.today().isoformat()` but can be set manually 
+        using an ISO formatted string (ex. "2019-10-31").
+3.  Run [serbo\_croat\_split.py](serbo_croat_split.py).
```
kylebgorman: Wonder if we should give this a name like `split_hbs.py` or something. Is there any other language that might need this?

languages/wikipron/src/serbo_croat_split.py
```diff
@@ -0,0 +1,53 @@
...
+
+import re
+import os
+import logging
+
+from codes import SERB_CROAT_PATH
+
+
+def _contains_latin(word):
+    return bool(re.search("[\u0000-\u007F]", word))
```
kylebgorman: I think this check is too weak. Here's what I'd suggest. 

`_all_latin` is true iff *every* character in a word is ASCII. 
`_all_cyrillic` is true iff *every* character in a word is Cyrillic (have to allow for dash, which can at least be used in some other Cyrillic scripts). 

If a word doesn't pass one or the other, log it and move on.

languages/wikipron/src/serbo_croat_split.py
```diff
@@ -0,0 +1,53 @@
...
+    phonetic = "phonetic.tsv"
+    phonemic = "phonemic.tsv"
+
+    try:
+        with open(f"{SERB_CROAT_PATH}{phonetic}") as serb_croat_data:
+            _split_file(SERB_CROAT_PATH, phonetic, serb_croat_data)
+        os.remove(f"{SERB_CROAT_PATH}{phonetic}")
+    except FileNotFoundError as err:
+        logging.info(
+            "No Serbo-Croatian phonetic tsv: %s", err,
```
kylebgorman: `TSV`

languages/wikipron/src/serbo_croat_split.py
```diff
@@ -0,0 +1,53 @@
...
+        )
+
+    try:
+        with open(f"{SERB_CROAT_PATH}{phonemic}") as serb_croat_data:
+            _split_file(SERB_CROAT_PATH, phonemic, serb_croat_data)
+        os.remove(f"{SERB_CROAT_PATH}{phonemic}")
+    except FileNotFoundError as err:
+        logging.info(
+            "No Serbo-Croatian phonemic tsv: %s", err,
```
kylebgorman: ditto

languages/wikipron/languages_summary.tsv
kylebgorman: You might also want it to read `hbs-latin_phonemic` (etc.) since that makes the intended interpretation a bit clearer.

languages/wikipron/src/README.md
lfashby: The only thing I've come across is that the [last](https://en.wiktionary.org/wiki/%E1%9C%80#Tagalog) Wiktionary Tagalog entry is written in Baybayin script, which we scrape and is included [in our Tagalog TSV](https://github.com/kylebgorman/wikipron/blob/master/languages/wikipron/tsv/tgl_phonemic.tsv#L1391).

It's possible other languages use two scripts, but I haven't come across them yet.

kylebgorman: I suggest making the file names `split_*.py` then so if there are more of them, they sort nicely.

lfashby:
Pushed these changes up now so we can talk about the `.encode()` stuff later.

kylebgorman:
Roger that---everything else looks good.

PR #106, opened 2019-11-24
Adds `_skip_pron` to test for the presence of '-'.

### LINKED ISSUE #105: [eng] dashed pronunciations
The English data contains roughly 300-400 entries where the pronunciation is given incompletely, with a dash at the beginning, middle, or end of the entry (usually indicating that we're supposed to fill in the rest of the pronunciation ourselves). Some examples:

    arabica ə - ɹ ă b ĭ - k ə
    observer    ɒ b -
    fornicatory - t o ʊ ɹ i

I would propose that we should remove all these punctuations and can do so on a language-independent basis.

kylebgorman:
Also adds associated test.

Fixes #105.

### DIFF
--- a/wikipron/extract/core.py
+++ b/wikipron/extract/core.py
@@ -13,10 +13,10 @@ if typing.TYPE_CHECKING:
 
 def _skip_pron(pron: str) -> bool:
     if "-" in pron:
-        return False
+        return True
     if " " in pron:
-        return False
-    return True
+        return True
+    return False
 
 
 def yield_pron(
--- a/wikipron/extract/core.py
+++ b/wikipron/extract/core.py
@@ -11,6 +11,14 @@ if typing.TYPE_CHECKING:
     from wikipron.typing import Iterator, Pron
 
 
+def _skip_pron(pron: str) -> bool:
+    if "-" in pron:
+        return False
+    if " " in pron:
+        return False
+    return True
+
+
 def yield_pron(
     request_html: requests_html.Element,
     ipa_xpath_selector: str,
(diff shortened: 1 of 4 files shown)


kylebgorman:
If accepted we should also regenerate English but I'll leave that to @lfashby

lfashby:
This should work. I wonder if we should add this logic to the `yield_pron()` function in `core.py`, or leave it in `scrape.py` next to `_skip_word()`. All our extraction functions use `yield_pron()` and skipping dashes could be slipped in after:
```
        # Skips examples with a space in the pron.
        if " " in pron:
            continue
```

kylebgorman:
"skipping space in prons" and "skipping dash in prons" should definitely go in the same place though I have no opinion about what space that ought to be. I just liked the rough paralleism of `_skip_word` and `_skip_pron`.

kylebgorman:
PTAL.

wikipron/extract/core.py
```diff
@@ -11,6 +11,14 @@
     from wikipron.typing import Iterator, Pron
 
 
+def _skip_pron(pron: str) -> bool:
+    if "-" in pron:
+        return False
+    if " " in pron:
+        return False
+    return True
```
jacksonllee: Are the boolean values are flipped? For instance, if there is a dash, we want to return `True` (not `False`). The tests also seem to be backwards.

kylebgorman:
Yes.

lfashby:
From looking at our data, it looks like the majority of dashes are of the `observer    ɒ b -` kind, where they are indicating alternate pronunciations of specific parts of a word, but I did find [one instance](https://en.wiktionary.org/wiki/camii) where it looks like a dash is being used as a syllable boundary. I don't think there is anything wrong with filtering those out as well, but just thought I'd mention it.

PR #109, opened 2019-11-26
Adds Japanese extraction function and post-processing.

### LINKED ISSUE #73: [jpn] Japanese support
Bringing over the discussion on Japanese from #62...

* Which script(s) should we target? Head words of Japanese entries on Wiktionary could be in Kana, Kanji, or a mix of both? Do we filter or not? Or do we want some sort of post-processing like what's being planned for Serbo-Croatian (#62)?
* In parallel with the WIkiPron package, there's [this prior work](https://github.com/kylebgorman/wikipron/blob/abecf0341268a9426e489226599ae8881c02aaf3/languages/jpn_wik/scrape.py) by Alan/Kyle for scraping Japanese specifically. Maybe incorporate this into WikiPron somehow?

@lfashby @kylebgorman Please edit or comment as you see fit.

lfashby:
This is my attempt to write an extraction function for Japanese. It isn't very pretty, but it seems to work. I've included the Japanese data that I scraped using this extraction function and a post-processing step for splitting the Japanese TSVs into hiragana and katakana. 

The names of the TSVs are `jpn_hira_...` and `jpn_kana_...` because "Hira" and "Kana" are the [ISO 15294 ](https://en.wikipedia.org/wiki/ISO_15924) designation for the scripts but I can change them if we'd like.

I can also add additional comments in `extract/jpn.py` if they would be helpful. 

I'm sure there will be suggestions to change certain things in `jpn.py`. I won't have too much free time until Wednesday night so I'll address the changes then. 

Resolves #73


kylebgorman:
Generally endorsing this though I would like Jackson to have the final word. 

I don't really see anything here such that it's obvious that it can be done better. And 15k entries is nothing to sneer at.

wikipron/extract/jpn.py
```diff
@@ -0,0 +1,229 @@
...
+      /a
+        [starts-with(@href, "#Etymology")]
+"""
+
+
+# Some pages may not have etymology section.
+def _check_etymologies(request: requests.Response):
+    count = 0
+    for a_element in request.html.xpath(_TOC_ETYMOLOGY_XPATH_SELECTOR):
+        count += 1
```
jacksonllee: Would `count = len(request.html.xpath(_TOC_ETYMOLOGY_XPATH_SELECTOR))` simply work? I can't remember if `request.html.xpath` already returns a materialized array of things that would allow `len`. If not, and if we're keeping this `for` loop, `a_element` could simply be `_` as it's not used.

wikipron/extract/jpn.py
```diff
@@ -0,0 +1,229 @@
...
+    try:
+        request.html.xpath(pron_path)[0]
+    except IndexError:
+        return
+
+    for upper_pron_ele in request.html.xpath(pron_path):
+        prons = []
+        for pron in yield_pron(upper_pron_ele, IPA_XPATH_SELECTOR, config):
+            prons.append(pron)
+        yield prons
```
jacksonllee: Would the following work?

```python
prons = list(yield_pron(upper_pron_ele, IPA_XPATH_SELECTOR, config))
yield prons
```

wikipron/extract/jpn.py
```diff
@@ -0,0 +1,229 @@
...
+) -> "Iterator[Pron]":
+    pron_path = _PRON_XPATH_SELECTOR.format(
+        word_to_work_from=word_target,
+        heading=_check_etymologies(request),
+        second_ul="",
+    )
+    for pron_element in request.html.xpath(pron_path):
+        prons = []
+        for pron in yield_pron(pron_element, IPA_XPATH_SELECTOR, config):
+            prons.append(pron)
```
jacksonllee: `prons = list(yield_pron(pron_element, IPA_XPATH_SELECTOR, config))` ?

jacksonllee:
Forgot to mention this: Should we remove `languages/jpn_wik/` and everything in this subdirectory?

kylebgorman:
Yes, IMO, that's now handled nicely.

languages/wikipron/README.md
```diff
@@ -91,8 +91,10 @@
 | [TSV](tsv/gle_phonetic.tsv) | gle | Irish | Irish | True | Phonetic | 1646 |
 | [TSV](tsv/ita_phonemic.tsv) | ita | Italian | Italian | True | Phonemic | 9612 |
 | [TSV](tsv/ita_phonetic.tsv) | ita | Italian | Italian | True | Phonetic | 5504 |
-| [TSV](tsv/jpn_phonemic.tsv) | jpn | Japanese | Japanese | False | Phonemic | 135 |
-| [TSV](tsv/jpn_phonetic.tsv) | jpn | Japanese | Japanese | False | Phonetic | 24746 |
+| [TSV](tsv/jpn_hira_phonemic.tsv) | jpn | Japanese | Japanese (Hiragana) | False | Phonemic | 137 |
+| [TSV](tsv/jpn_hira_phonetic.tsv) | jpn | Japanese | Japanese (Hiragana) | False | Phonetic | 14494 |
+| [TSV](tsv/jpn_kana_phonemic.tsv) | jpn | Japanese | Japanese (Katakana) | False | Phonemic | 1 |
```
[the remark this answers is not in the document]
kylebgorman: +1 we normally do this and it's mentioned in the README change

wikipron/extract/jpn.py
```diff
@@ -0,0 +1,229 @@
...
+
+def _combine_word_pron_pairs(
+    words: "Iterator[Word]", prons: "Iterator[Pron]"
+) -> "Iterator[WordPronPair]":
+    for word in words:
+        try:
+            associated_prons = next(prons)
+        except StopIteration:
+            pass
+        yield from (zip(itertools.repeat(word), associated_prons))
```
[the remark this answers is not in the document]
kylebgorman: +1

wikipron/extract/jpn.py
kylebgorman: If it's not an actual `len`-able object the hack is:

    count = sum(1 for _ in request.html.xpath(...))

wikipron/extract/jpn.py
kylebgorman: can't you do `yield from yield_pron(...)`?

kylebgorman:
> Forgot to mention this: Should we remove `languages/jpn_wik/` and everything in this subdirectory?

Yes please, though it doesn't have to be done in this PR if you don't feel like it.

wikipron/extract/jpn.py
jacksonllee: IIUC, the code was trying to yield a list of prons, rather than yield the prons one by one, and hence my suggested refactoring. That being said, now that I'm reading this again in the larger context, I think Lucas was trying to generate both "lower" and "upper" prons, chain them together, and then yield the prons one at a time. In this case, I think there's a more natural way to refactor the code. The first part is what Kyle has suggested: We should do `yield from yield_pron(...)` here, so that `_yield_jpn_upper_pron` yields the prons one by one. Other parts are in my updated comments below.

wikipron/extract/jpn.py
```diff
@@ -0,0 +1,229 @@
...
+            # etymologies are exceedingly rare. I'm not sure I've seen
+            # any.
+            prons += next(upper_prons)
+        except StopIteration:
+            pass
+        # Yielding here because we don't want to collect all pronunciation
+        # entries on a page at the same time. Doing so would make it difficult
+        # to connect words to their prons appropriately. We only want to grab
+        # the prons that are linked to the word we have scraped.
+        yield prons
```
jacksonllee: We probably don't have to catch `StopIteration`, if we use `itertools.chain`:

```python
for pron_element in request.html.xpath(pron_path):
    lower_prons = yield_pron(...)
    upper_prons = _yield_jpn_upper_pron(...)
    # The following won't throw an error even if either or both of lower_prons and upper_prons are empty
    prons = itertools.chain(lower_prons, upper_prons)
    yield from prons
```

wikipron/extract/jpn.py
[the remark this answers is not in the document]
jacksonllee: If `prons` returned by `_yield_jpn_lower_pron` yields prons one by one, then we don't need to catch `StopIteration` here:

```python
for word in words:
    yield from zip(itertools.repeat(word), prons)
```

wikipron/extract/jpn.py
lfashby: I like the `itertools.chain()` approach, but I'm hesitant to exhaustively run `_yield_jpn_upper_pron(...)` each iteration of the loop. Then it would collect all the upper prons from all the Etymologies each time we grab the lower pron or prons for a specific word, potentially leading to a mismatch of words and pronunciations. (I say potentially because I've never seen any page with multiple etymologies that have multiple <ul>'s.)

languages/wikipron/README.md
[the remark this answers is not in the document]
lfashby: My bad, I changed `split_jpn` to automatically remove the small files (and should probably add that behavior to `split_hbs` just to be consistent), but forgot to re-run `generate_summary.py`.

wikipron/extract/jpn.py
lfashby: The idea in both of these pron yielding functions is that I want to exhaustively run `yield_pron()` on each iteration of the upper for loops in each of the pron yielding functions. 

When the next `next(prons)` in `_combine_word_pron_pairs()` gets called, the upper for loop in both pron yielding functions gets ‘incremented’ (if that’s the right word?) and we target the subsequent word-pron pair group on the page and grab the prons in that group.

I am trying to potentially yield a list of prons in each function. Jackson's original `list()` suggestion is a simplified way of exhaustively running `yield_pron()`, but it still leaves this:
```
        try:
            prons += next(upper_prons)
        except StopIteration:
            pass
```
block in `_yield_jpn_lower_pron`. (If this comment and the comment I left on the `itertools.chain()` thread don't make any sense then let me know.)

lfashby:
I addressed most of the suggestions and added some comments to `jpn.py`. If there are any ideas on how to refactor the pron yielding functions in light of my responses to your comments then let me know. I'll be out for the rest of the day but will be around tomorrow.

lfashby:
Great. I learned a lot while working on this so I think it would be rewarding to try and rework it over winter break when I have some free time.

PR #110, opened 2019-11-26
Welsh dialect support

### LINKED ISSUE #108: [cym] Add dialect labels
It looks like there are "North Wales" and "South Wales" entries.

kylebgorman:
Splits Welsh into North Wales and South Wales dialects. It appears nearly every entry in Wiktionary has both dialectal forms and they end to have minor vowel differences. This massively reduces the number of prons per word in Welsh.

Closes #108.


kylebgorman:
> LGTM. It look like you ran `remove_duplicates.sh` as well. Did you read the "big scrape" readme instructions?

Yeah, that's what I'm supposed to do, right? Steps to do this:

* remove the old Welsh data
* make a copy of `languages.json` that just has the Welsh dialects
* run `./scrape.py`
* run `./remove_duplicates.sh`
* reset `languages.json` and then add the Welsh dialect bit
* run `generate_summary.py`

lfashby:
> Yeah, that's what I'm supposed to do, right?

Yes, exactly. I was just wondering if the new section of the big scrape readme was helpful. (It doesn't specifically cover adding dialects for a language but the steps are similar.)

kylebgorman:
Yeah, I think so!

PR #111, opened 2019-11-27
Updated English and Latin data

lfashby:
Re-ran English to get hyphen-free data. Interestingly there are also `~` characters in [certain English transcriptions](https://en.wiktionary.org/wiki/path#Etymology_1). They seem to serve a similar purpose as the hyphens in that they indicate alternate pronunciations. We end up with data like this:
```
path 	pʰ æ θ ~ pʰ ɛ ə θ ~ pʰ e ə θ
```

Re-ran Latin because of the changes made to casefolding.


lfashby:
The English data also differs from previous English data in that it is parsed with `segments`.

I'll restart Russian soon if there aren't any smaller languages to re-run.

PR #115, opened 2019-11-30
Updated Russian data

lfashby:
This is a complete scrape of Russian. When I ran `remove_duplicates.sh` it rearranged some entries in the new Welsh TSVs (which Kyle sorted on his computer). This is not too dissimilar from something I noticed earlier when I ran `remove_duplicates.sh` on a different computer and saw that it rearranged files that had already been sorted (and considered certain things duplicates that were not actually duplicates - which did not happen in this instance). 

The suggestion at the time of using `usr/bin/sort` as opposed to just `sort` in `remove_duplicates.sh` did not change this behavior. Should I open a ticket regarding this?


jacksonllee:
> The suggestion at the time of using usr/bin/sort as opposed to just sort in remove_duplicates.sh did not change this behavior. Should I open a ticket regarding this?

+1 for opening a ticket. It looks like the sorting changes in this PR treated the tab to be "less" than the letters, which seems right based on the ASCII table. I wonder if we are on different systems (e.g., MacOS vs Linux) where `sort` may behave differently by any chance? Not sure.

aryamccarthy:
Does this mean we have to rerun Russian experiments on the new data?

-am

kylebgorman:
I didn't use Russian for the validation experiments for the simple reason
that it wasn't ready, so no we're good.

PR #117, opened 2019-12-02
Adds ToC and data section to README

kylebgorman:
Jackson, if this is roughly on the right track, I'd suggest you can add the versioning links later...just wanted to get a draft of this out there.


README.md
```diff
@@ -1,104 +1,127 @@
...
+Python API
+----------
 
-The underlying module can also be used from Python.
-A standard workflow looks like:
+The underlying module can also be used from Python. A standard workflow looks
+like:
 
-```python
+``` {.python}
```
jacksonllee: It looks like this line has to stay as-is for syntax highlighting to work. I was looking [here](https://github.com/kylebgorman/wikipron/tree/readme) to see how the README would look after this PR, where the Python code highlighting didn't show up.

kylebgorman: I just went through and manually reverted all of those. I'm using Pandoc to reflow.

kylebgorman:
PTAL

PR #122, opened 2019-12-15
Integrates Wortschatz frequencies

lfashby:
This PR adds two scripts and a json file to a new directory `src/frequencies`. `grab_wortschatz_data.py` automatically downloads and unpacks [Wortschatz tars](http://wortschatz.uni-leipzig.de/en/download/) and `merge.py` merges in the Wortschatz word frequency counts with our TSVs such that we end up with three column TSVs (our old TSVs are replaced):
```
алтын	ɑ ɫ t ɯ̞ n	745
```
This is currently done for 61 languages, I'll see if I can add more before our next big scrape.

For words that we scrape and that Wortschatz doesn't have a frequency count for, the third column is set to 0.

I added the `tars` and `freq_tsvs` directories (both created when running `grab_wortschatz_data.py`) to `.gitignore` as opposed to automatically deleting them once `merge.py` completes because we download about 10 gigabytes worth of Wortschatz stuff and it's a pain to have to re-download it all every time we want to test these two scripts.


languages/wikipron/src/frequencies/grab_wortschatz_data.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python
+
+import json
+import logging
+import os
+import requests
+import tarfile
+import time
+
+with open("wortschatz_languages.json", "r") as langs:
```
kylebgorman: why is this at global scope?

kylebgorman:
Basically looks good to me. I don't understand the *.tar.gz in the .gitignore though. I get why you want to keep the tarballs around, but maybe just add the directory as a .gitignore, not all filetypes like that?

languages/wikipron/src/frequencies/grab_wortschatz_data.py
lfashby: Good question!

lfashby:
I experimented with different ways of grabbing the Wortschatz data (including with cURL) to see if I could resolve the 404 response problem, but generally got the same behavior. The solution in `grab_wortschatz_data.py` is not ideal but it does at least eventually collect all the tarballs we need and only needs to be run once.

languages/wikipron/src/frequencies/grab_wortschatz_data.py
```diff
@@ -51,6 +48,9 @@ def unpack():
 
 
 def main():
+    with open("wortschatz_languages.json", "r") as langs:
```
kylebgorman: make this string a global variable at the top?

lfashby: Or perhaps I should add it to the other path constants in `src/codes.py`?

kylebgorman:
That'd be fine too, whatever you think is more consistent.

PR #123, opened 2020-01-08
Generalized split and README expansion

### LINKED ISSUE #121: [aze] Azerbaijani has three 3 scripts
Our Azerbaijani data - [phonetic](https://github.com/kylebgorman/wikipron/blob/master/languages/wikipron/tsv/aze_phonetic.tsv), [phonemic](https://github.com/kylebgorman/wikipron/blob/master/languages/wikipron/tsv/aze_phonemic.tsv) - contains Latin, Cyrillic, and Arabic scripts.

lfashby:
This PR adds a new script `generalized_split.py` to `wikipron/src` that replaces the old `split_jpn.py` and `split_hbs.py` scripts while retaining their functionality and handling splitting Azerbaijani. The aim of `generalized_split.py` is to be able to handle any language with multiple scripts that we might want to split. To do so, it uses the `regex` module's Unicode script property. (I have not yet added a `requirements.txt` for the scraping scripts because I had a few questions about that.) 

`generalized_split.py` is called within the newly renamed `remove_duplicates_and_split.sh` as this seemed like an appropriate place for it and reduces the number or steps involved in running the big scrape. I've updated the big scrape readme to reflect this change.

This PR also moves the `frequencies` directory from a subdirectory within `src` to sister with `src` and adds a brief README to `frequencies`. I have not added steps for running the frequencies scripts as there are only two of them and the description of them in the README makes it clear enough how they should be run.

A few concerns:
- The `regex` method of splitting the TSVs seems to work well, but it does reduce the number of Japanese Katakana phonetic entries by 2000. It'd be nice to have a Japanese speaker look over the two different version of our Katakana data and see if they can identify why the `regex` split is more restrictive than our previous split. (I can do that when school starts back up, but if you have any intuitions let me know.)
- `generalized_split.py` is called within `remove_duplicates_and_split.sh` while `remove_duplicates_and_split.sh` is iterating through the TSVs in our `tsv` directory. I was surprised to find that despite `generalized_split.py` adding and removing TSVs to the directory that is being iterated through, no errors are thrown and nothing seems to go wrong. Am I just lucky that no errors are thrown (as I believe they would be if `remove_duplicates_and_split.sh` was a Python script) and is it dangerous to add and delete files in a directory while iterating through that directory with Bash?

Resolves #121


kylebgorman:
@alanwongs can look at the Japanese data for us, perhaps; I've added him if he's willing to help out.

I would run those two scripts in serial, not parallel, if you can even imagine a case where it might be bad for them to run in parallel.

languages/wikipron/frequencies/README.md
```diff
@@ -0,0 +1,26 @@
+Frequencies scripts
+===================
+
+The scripts in this directory are responsible for downloading
+word frequency counts from the
+[Leipzig Corpora Collection](https://wortschatz.uni-leipzig.de/en/download/)
+and merging those counts into [our corresponding TSVs](../tsv/).
+[grab\_wortschatz\_data.py](grab_wortschatz_data.py) downloads
+and unpacks the tars provided by the aforementioned Corpora Collection.
```
kylebgorman: `TARs` cf. `TSVs`---short for "tape archive" originally

languages/wikipron/src/remove_duplicates_and_split.sh
```diff
@@ -5,4 +5,5 @@ for TSV in ../tsv/*.tsv; do
     # Explicitly uses byte-wise comparison for sorting
     # rather than a locale-dependent comparison.
     LC_ALL=C sort -u -o "${TSV}" "${TSV}"
+    python generalized_split.py $TSV
```
kylebgorman: execute `chmod +x generalized_split.py` (if you haven't already), make sure it has the appropriate hashbang at the top, then just do `./generalized_split.py`.

Also do `"${TSV}"` instead of `$TSV`, which makes this robust to the presence of space or other escape characters in the filename path.

languages/wikipron/src/generalized_split.py
```diff
@@ -0,0 +1,51 @@
+import json
+import os
+import regex
```
kylebgorman: Need to add this to `requirements.txt` before submitting---we can discuss offline?

alanwongs:
The changes seem to have stripped away the katakana words with the vowel lengthening mark "ー" in them (just under 1400 words).

kylebgorman:
Huh, so the Unicode character class isn't quite broad enough in this case.

Let me suggest that we just treat this as a special case such that if the
character class is "Katakana" we also union it with this length marker?
Would that work?

lfashby:
Thanks for the help Alan. 

I suppose a union would work though I'd like to poke around for other solutions as well. There is some discussion of this vowel lengthening mark online so I'd imagine others have had this problem. I'll be out this evening but will see if I can implement a solution tomorrow.

lfashby:
These changes pick up roughly another 1000 Katakana entries. I realize now I need to re-run the frequencies scripts if we don't want Japanese to lose its word counts in this PR. Is that worth doing now or can we wait until the next big scrape?

kylebgorman:
Can wait, IMO.

lfashby:
Let me know if there are any problems with the `requirements.txt` I put together. I think I'll try and kick off another scrape either today or tomorrow once this gets accepted.

languages/wikipron/requirements.txt
```diff
@@ -0,0 +1,5 @@
+flake8==3.7.9
```
kylebgorman: This is fine. Maybe we want to remove the versioning from requests and requests-html under the assumption they're probably stable but let's get this in and tinker later.

Dumb question also---why do we need flake8 here?

lfashby: I suppose we don't need flake8 here, but I've added it because I run it prior to committing.

kylebgorman: Yeah, I'd cut it. Ready to merge otherwise?

lfashby: I believe so. I'll clean up `requirements.txt` in a future PR.

PR #124, opened 2020-01-22
Extraction function for Chinese

### LINKED ISSUE #72: [chi] Chinese support
During the big scrape, wikipron worked on scraping Chinese for about 24 hours and the tsvs generated only contained a few entries. 

[Chinese](https://en.wiktionary.org/wiki/%E9%87%91%E6%9F%91#Chinese) entries on Wiktionary appear to have IPA information in `<li>` elements within a collapsible `<div>`. As mentioned by Jackson [here](https://github.com/kylebgorman/wikipron/issues/57#issuecomment-539296420), these entries often contain IPA transcriptions for various Chinese languages and dialects within those languages. Mandarin IPA transcriptions are often given in Sinological IPA, which may contain unofficial(?) IPA symbols.

These entries are perhaps causing problems for `_LI_SELECTOR_TEMPLATE` in `config.py` because of intervening `<small>` elements between the `<sup>` and `<span class="IPA">` that we want to target?

([Certain](https://en.wiktionary.org/wiki/%E9%9F%A6%E4%BC%AF#Chinese) Chinese entries will simply link to other entries.)

lfashby:
This PR adds an extraction function for Chinese that targets Mandarin pronunciations. It also filters the Mandarin data we collect by removing everything that is not part of the Han Unicode script leaving us with a TSV `chi_hani_phonemic.tsv` that contains 117991 entries. 'hani' is used in order to follow the ISO 15924 conventions we observe when naming TSVs that have been split, but since we are only filtering out Latin entries (and some emojis) and aren't splitting the Mandarin data into multiple TSVs perhaps it would make more sense to leave it as `chi_phonemic.tsv`?
The method used for filtering the Mandarin data could be applied to other languages, like Korean, that we want to remove Latin entries from.  

This PR also fixes a problem in `merge.py` and some previously identified problems with the `requirements.txt` for the scraping scripts.

I think this resolves #72 although there is some other Chinese language data to be grabbed from scrapping `chi` like Min Nan and Cantonese transcriptions as mentioned in #57 .


lfashby:
I forgot to mention that I modified `extract/core.py` to not skip prons with spaces in them when scrapping Chinese because the majority of Chinese prons contain spaces. This has required me to modify `test_extract.py` because I'm now passing our config to the `skip_pron` function. If there are more elegant ways of handling the spaces problem with Chinese or if the changes made to `test_extract.py` are not appropriate, please let me know.

kylebgorman:
Adding @jacksonllee in case he has any thoughts.

lfashby:
> This looks fine and like it closes #72 but there are huge number of other languages' files (from the ongoing scrape) affected; is that intended?

Good point, I hadn't thought of that. I reverted those changes. If we're desperate to add word frequency counts for the Chinese data I can do so, but it'll be easier to add them at the end of the current scrape.

lfashby:
> If we're handling Mandarin Chinese specifically, the ISO 639-3 language code should be `cmn`, not `chi`.

Hi Jackson,

That's a good point. The `iso639` package we use doesn't recognize `cmn` so we could add it to `languagecodes.py` and redirect it to the [Chinese](https://en.wiktionary.org/wiki/Category:Chinese_terms_with_IPA_pronunciation) Wiktionary page with `cmn: "Chinese"`. Do you think the same thing should be done with Cantonese? (`yue` is also not recognized by the `iso639` package.). I believe that is what you were suggesting in #57.

These changes would mean a user would still be able to run `wikipron chi`, but they wouldn't get any data. I assume it would not be useful if `chi` grabbed all the pronunciations from the various languages for each word?

jacksonllee:
Hi Lucas, I think what you've said is very much in line with what I have in mind. Just to be explicit --

> The iso639 package we use doesn't recognize cmn so we could add it to languagecodes.py and redirect it to the Chinese Wiktionary page with cmn: "Chinese". 

Yes, I think this is the right approach given what we've got.

> Do you think the same thing should be done with Cantonese?

Yes. Cantonese as `yue` is pretty much a parallel story with Mandarin as `cmn`, to the extent that Cantonese, Mandarin, as well as other Chinese varieties are scrape-able from the Wiktionary's [Chinese](https://en.wiktionary.org/wiki/Category:Chinese_terms_with_IPA_pronunciation) as described in my comment at https://github.com/kylebgorman/wikipron/issues/57#issuecomment-539296420.

> These changes would mean a user would still be able to run wikipron chi, but they wouldn't get any data. I assume it would not be useful if chi grabbed all the pronunciations from the various languages for each word?

Exactly. Running `wikipron chi` should return nothing useful.

lfashby:
I made the changes requested. When we go to write a Cantonese extraction function we may need to modify `extract/__init__.py` in some way (or modify and rename `cmn.py` to handle Mandarin and Cantonese). We want both `cmn` and `yue` to target the "Chinese" Wiktionary language and `extract/__init__.py` will not be able to assign a different extraction function to `yue` if it has the key as `cmn`.

wikipron/extract/__init__.py
```diff
@@ -2,6 +2,7 @@
 from wikipron.extract.lat import extract_word_pron_latin
 from wikipron.extract.tha import extract_word_pron_thai
 from wikipron.extract.jpn import extract_word_pron_jpn
+from wikipron.extract.cmn import extract_word_pron_cmn
```
kylebgorman: Please sort these imports alphabetically.

wikipron/extract/__init__.py
```diff
@@ -11,4 +12,5 @@
     "Latin": extract_word_pron_latin,
     "Thai": extract_word_pron_thai,
     "Japanese": extract_word_pron_jpn,
+    "Chinese": extract_word_pron_cmn,
```
kylebgorman: Ditto, please sort.

PR #131, opened 2020-03-03
Prep for v1.1.0 release

jacksonllee:
As soon as this is merged, I'll make a new release on both PyPI and GitHub.


jacksonllee:
Question: I think we should add the citation to our forthcoming LREC paper. Thoughts?

kylebgorman:
> Question: I think we should add the citation to our forthcoming LREC paper. Thoughts?

Yes please.

kylebgorman:
Agree.

PR #132, opened 2020-03-21
Increased coverage of language codes not in iso639 lib

tpimentelms:
This PR adds new language codes to the dictionary in `wikipron/languagecodes.py` to increase language coverage.

### DIFF
--- a/wikipron/languagecodes.py
+++ b/wikipron/languagecodes.py
@@ -259,5 +259,7 @@ LANGUAGE_CODES = {
     # Mandarin Chinese. ISO 639-3 only.
     "cmn": "Chinese",
     # Abkhaz. Would be Abkhazian in ISO 639.
-    "abk": "Abkhaz"
+    "abk": "Abkhaz",
+    # Avar. Would be Avaric in ISO 639.
+    "ava": "Avar"
 }
--- a/wikipron/languagecodes.py
+++ b/wikipron/languagecodes.py
@@ -261,5 +261,7 @@ LANGUAGE_CODES = {
     # Abkhaz. Would be Abkhazian in ISO 639.
     "abk": "Abkhaz",
     # Avar. Would be Avaric in ISO 639.
-    "ava": "Avar"
+    "ava": "Avar",
+    # Buryat. Would be Buriat in ISO 639.
+    "bua": "Buryat",
 }
--- a/wikipron/languagecodes.py
+++ b/wikipron/languagecodes.py
@@ -264,4 +264,6 @@ LANGUAGE_CODES = {
     "ava": "Avar",
     # Buryat. Would be Buriat in ISO 639.
     "bua": "Buryat",
+    # Chukchi. Would be Chukot in ISO 639.
+    "ckt": "Chukchi",
 }
--- a/wikipron/languagecodes.py
+++ b/wikipron/languagecodes.py
@@ -253,7 +253,7 @@ LANGUAGE_CODES = {
     # Bouyei. ISO 639-3 only.
     "pcc": "Bouyei",
     # Lamboya. ISO 639-3 only.
-    "lmy": "Lamboya",
+    "lmy": "Laboya",
     # Moroccan Arabic. ISO 639-3 only.
     "ary": "Moroccan Arabic",
     # Mandarin Chinese. ISO 639-3 only.
@@ -268,8 +268,6 @@ LANGUAGE_CODES = {
     "ckt": "Chukchi",
     # Burushaski. ISO 639-3 only.
     "bsk": "Burushaski",
-    # Galician. Not in iso639 lib.
-    "glc": "Galician",
     # Evenki. Not in iso639 lib.
     "evn": "Evenki",
     # Southern Yukaghir. Not in iso639 lib.
--- a/wikipron/languagecodes.py
+++ b/wikipron/languagecodes.py
@@ -298,4 +298,16 @@ LANGUAGE_CODES = {
     "koi": "Komi-Permyak",
     # Komi-Zyrian. Not in iso639 lib.
     "kpv": "Komi-Zyrian",
+    # Lak. Not in iso639 lib.
+    "lbe": "Lak",
+    # Lezgi. Would be Lezghian in ISO 639.
+    "lez": "Lezgi",
+    # Eastern Mari. Not in iso639 lib.
+    "mhr": "Eastern Mari",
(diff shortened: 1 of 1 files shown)


kylebgorman:
Hi! @jacksonllee tells me that there are some issues with this he noticed; for instance we already scrape Galician as `glg`; `glc` is Bon Gula. I'm going to assign him and @lfashby as reviewers since I don't have a chance to look more closely right now.

wikipron/languagecodes.py
```diff
@@ -257,5 +257,57 @@
...
+    # Avar. Would be Avaric in ISO 639.
+    "ava": "Avar",
+    # Buryat. Would be Buriat in ISO 639.
+    "bua": "Buryat",
+    # Chukchi. Would be Chukot in ISO 639.
+    "ckt": "Chukchi",
+    # Burushaski. ISO 639-3 only.
+    "bsk": "Burushaski",
+    # Galician. Not in iso639 lib.
+    "glc": "Galician",
```
jacksonllee: One issue I've noticed: `glc` is for Bon Gula, not Galician. For Galician, the ISO 639-2 or -3 code is `glg`. FWIW, we've already scraped the Galician data from Wikitionary, see [here](https://github.com/kylebgorman/wikipron/tree/7ea49af5802a60208382fcac17d5855dae56376f/languages/wikipron).

lfashby:
Regarding Jackson's comments, I'd like to add that Wikipron can currently scrape the languages that fail `test_languagecodes.py` but we still receive warnings about them because we did not add the lowercase Wiktionary language name in addition to the iso639 code. Something like:
```
    # Bouyei. ISO 639-3 only.
    "pcc": "Bouyei",
    "bouyei": "Bouyei",
```
would prevent the warning for these languages. Additionally, `"lmy": "Lamboya",` should be changed to `"lmy": "Laboya", as "Lamboya" is the iso639 language name. 

One question that the proposed changes brings up is whether we should expand `languagecodes.py` to handle someone using a code like `ekk` for Estonian. `est` apparently refers to the Estonian macrolanguage while `ekk` refers to Standard Estonian.

wikipron/languagecodes.py
tpimentelms: Oh, my bad. Will remove this now.

tpimentelms:
But couldn't these language codes be there, but not be automatically scraped by wikipron? 
I added these ones because I needed them for a personal project.

About the `ekk` vs `est` discussion I don't have an opinion. I can remove it if you think that is better.

lfashby:
I don't think there is any harm in having these in `languagecodes.py`. Looks to me like all the language codes are correct and running a Wikipron scrape with them doesn't throw any errors. I'd like to hear what Jackson has to say about it though, perhaps if we include these it would make sense to eventually remove the min language size requirement in `test_languagecodes.py`?

Similar to the `ekk` and `est` thing there is also `per` versus `pes` for Persian - `per` is the iso639-2/B code for the Persian macro-language while `pes` is the iso639-3 code for Iranian Persian. Not sure which we prefer and I'd defer to Jackson on that question.

Just out of curiosity - did you use Wikipron to scrape for these languages and get usable data for all of them?

wikipron/languagecodes.py
```diff
@@ -252,10 +252,89 @@
...
+    "persian": "Persian",
+    # Greenlandic. Would be Kalaallisut in ISO 639.
+    "kal": "Greenlandic",
+    "greenlandic": "Greenlandic",
+    # Khanty. Not in iso639 lib.
+    "kca": "Khanty",
+    "khanty": "Khanty",
+    # Ket. Not in iso639 lib.
+    "ket": "Ket",
+    "ket": "Ket",
```
lfashby: I suppose when the name and the code are the same we can break from procedure and leave out the lowercase Wiktionary language name.

tpimentelms: It makes sense, I didn't pay attention to this when duplicating the lines.

tpimentelms:
I focused on getting only phonemic data (and not phonetic) and languages `'kat', 'ket', 'olo', 'kca', 'lbe', 'nio', 'hye'` didn't have any entries.
The others I fetched with wikipron. Didn't check the resulting files for the other languages yet, though.

wikipron/languagecodes.py
```diff
@@ -252,10 +252,88 @@
...
+    "kildin sami": "Kildin Sami",
+    # Northern Yukaghir. Not in iso639 lib.
+    "ykg": "Northern Yukaghir",
+    "northern yukaghir": "Northern Yukaghir",
+    # Nanai. Not in iso639 lib.
+    "gld": "Nanai",
+    "nanai": "Nanai",
+    # Persian. Not in iso639 lib.
+    "pes": "Persian",
+    "persian": "Persian",
```
jacksonllee, line 306: My spot checking tells me that some Persian entries have the dialect indication (e.g., "Iranian Persian" in [this entry](https://en.wiktionary.org/wiki/%D8%A2%D9%84%D9%88%D8%AF%DA%AF%DB%8C#Persian)), which suggests that we should use the macro-language code `per` here (and a WikiPron user who would like to target "Iranian Persian" specifically will have to specify the dialect flag for that).

To not block this PR further, I'd be fine with merging this PR as-is, and then the WikiPron developers can fix this ourselves.

wikipron/languagecodes.py
```diff
@@ -252,10 +252,88 @@
...
+    "evn": "Evenki",
+    "evenki": "Evenki",
+    # Southern Yukaghir. Not in iso639 lib.
+    "yux": "Southern Yukaghir",
+    "southern yukaghir": "Southern Yukaghir",
+    # Tundra Nenets. Not in iso639 lib.
+    "yrk": "Tundra Nenets",
+    "tundra nenets": "Tundra Nenets",
+    # Estonian. Not in iso639 lib.
+    "ekk": "Estonian",
```
jacksonllee, line 290: `ekk` looks fine to me here. The "iso639" package can already resolve `est` as Estonian, probably because `est` is in ISO 639-2 as well. `ekk` is ISO 639-3 only -- thanks for adding it!

It seems fine to point both `ekk` and `est` to "Estonian" for our purposes. I haven't found [Estonian entries on Wiktionary](https://en.wiktionary.org/wiki/Category:Estonian_terms_with_IPA_pronunciation) that would have a dialect label for "Standard Estonian" or something. If we learn more about Estonian (broadly construed) in the future and we decide to update the mapping, we'll certainly do so.

Issue #91, opened 2019-11-08
Add a --no-tone flag


jacksonllee:
Thai (#71) was handled by #90, but we'd still like a `--no-tone` flag to optionally remove the IPA tones.

kylebgorman:
+1. Wonder if this should default to false?

jacksonllee:
> Wonder if this should default to false?

Yes, it'd be the same treatment like `--no-stress` and `--no-syllable-boundaries` that we already have (so with `action="store_true"` in the CLI argparse set-up).

In general, I think our API design should be "if the user wants to _remove_ information, they should be forced to do something, e.g., explicitly applying a flag to say so".

jacksonllee:
Also, a note about what "tone" means for our purposes --

1. The IPA tone letters like `˧`, as used for the Thai entries.
2. The Chao tone numbers like `⁵`, as used for Chinese languages (cf. #72, #57).
3. The accents, as widely used for Bantu languages ([example from Chichewa](https://en.wiktionary.org/wiki/beseni#Chichewa)).

The `--no-tone` flag for removing tone would have to handle all these cases.

kylebgorman:
Super cool that the IPA decided not to standardize any of those! ;)

lfashby:
Looking at Jackson's Chichewa example in the comment above I'm wondering if Chichewa is like Latin or like Serbo-Croatian. The headwords often don't have tone diacritics, but the words on each entry page do.

Here is a line from the [about Chichewa](https://en.wiktionary.org/wiki/Wiktionary:About_Chichewa) page that makes me think it is like Serbo-Croatian:
> Entries when linked to or mentioned should not show tone unless there is reason to do so (for example, to distinguish words whose only difference is tone).

PR #135, opened 2020-03-28
Adds --no-skip-spaces-ortho and --no-skip-spaces-pron flags

### LINKED ISSUE #133: [vie] Enhancements
I notice a few obvious problems with Vietnamese:

1. As far as I can tell, nearly all Vietnamese words have three phonetic pronunciations: "Hà Nội", "Huế, and "Hồ Chí Minh City". We should just add dialects to `languages.json`
2. As is well-known, the Vietnamese Roman orthography puts a space between every syllable. So our data set is really only monosyllabic Vietnamese words! We could solve this by adding a flag (`--no-skip-space`) [here](https://github.com/kylebgorman/wikipron/blob/7ea49af5802a60208382fcac17d5855dae56376f/wikipron/scrape.py#L17), or we could make a language specific-extractor, I suppose.

lfashby:

`--no-skip-spaces-ortho` when set to True will not skip words that contain spaces (whether regular or non-breaking  - though I haven't come across any of these yet). This required modifying `cli.py`, `config.py`, and `wikipron/scrape.py`
`--no-skip-spaces-pron` when set to True will not skip transcriptions that contain spaces (whether regular or non-breaking). This required modifying `cli.py`, `config.py`, and `extract/core.py`. `extract/core.py` now also checks whether this flag is set and removes the `"#"` that the `segments` package inserts. (Is this the desired behavior discussed in #133?)

I also modified `languages.json` to include the appropriate flags for the relevant languages (Chinese, Vietnamese, Persian, Tibetan). `src/scrape.py` was slightly modified to make it easier to build the config dictionary passed to Wikipron. 
(It's my first time adding a flag so, though this seems to work, I'm not entirely sure I followed best practices, in particular for modifying `config.py`.)

### DIFF
--- a/languages/wikipron/src/scrape.py
+++ b/languages/wikipron/src/scrape.py
@@ -62,17 +62,15 @@ def main():
         languages = json.load(source)
     # "2020-01-15" (Big Scrape 3)
     cut_off_date = datetime.date.today().isoformat()
+    valid_keys = ["casefold", "no_skip_spaces_pron", "no_skip_spaces_ortho", "dialect"]
     for iso639_code in languages:
-        # config_settings = {
-        #     "key": iso639_code,
-        #     "casefold": languages[iso639_code]["casefold"],
-        #     "no_stress": True,
-        #     "no_syllable_boundaries": True,
-        #     "cut_off_date": cut_off_date,
-        # }
+        wikipron_accepted_settings = {valid_key: languages[iso639_code][valid_key] for valid_key in valid_keys if valid_key in languages[iso639_code]}
         config_settings = {
-            **languages[iso639_code],
+            "key": iso639_code,
+            "no_stress": True,
+            "no_syllable_boundaries": True,
             "cut_off_date": cut_off_date,
+            **wikipron_accepted_settings
         }
         if "dialect" not in languages[iso639_code]:
             _build_scraping_config(
--- a/languages/wikipron/src/scrape.py
+++ b/languages/wikipron/src/scrape.py
@@ -43,9 +43,7 @@ def _call_scrape(lang_settings, config, tsv_path):
     os.remove(tsv_path)
 
 
-def _build_scraping_config(
-    config_settings, wiki_name, dialect_suffix=""
-):
+def _build_scraping_config(config_settings, wiki_name, dialect_suffix=""):
     path_affix = f'../tsv/{config_settings["key"]}_{dialect_suffix}'
 
     phonemic_config = wikipron.Config(**config_settings)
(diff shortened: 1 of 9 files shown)


languages/wikipron/src/scrape.py
```diff
@@ -62,13 +60,19 @@ def main():
         languages = json.load(source)
     # "2020-01-15" (Big Scrape 3)
     cut_off_date = datetime.date.today().isoformat()
+    valid_keys = ["casefold", "no_skip_spaces_pron", "no_skip_spaces_ortho"]
     for iso639_code in languages:
+        wikipron_accepted_settings = {
```
kylebgorman: This makes the same lookup (`languages[iso639]`) twice. Can you rewrite this as a loop to avoid that?

wikipron/cli.py
```diff
@@ -74,6 +74,16 @@ def _get_cli_args(args: List[str]) -> argparse.Namespace:
             "To disable such IPA segmentation, apply this flag."
         ),
     )
+    parser.add_argument(
```
kylebgorman: This flag (and throughout) should probably be called "--no-skip-spaces-word" since we use the term `word` and `pron` throughout.

wikipron/extract/core.py
```diff
@@ -15,7 +15,7 @@
 def _skip_pron(pron: str, config: "Config") -> bool:
     if "-" in pron:
         return True
-    if " " in pron and config.language != "Chinese":
+    if (" " in pron or "\u00A0" in pron) and not config.no_skip_spaces_pron:
```
kylebgorman: Check the config option before doing the more expensive check for space.

wikipron/scrape.py
```diff
@@ -14,9 +14,9 @@
 _PAGE_TEMPLATE = "https://en.wiktionary.org/wiki/{word}"
 
 
-def _skip_word(word: str) -> bool:
+def _skip_word(word: str, no_skip_spaces: bool) -> bool:
     # Skips multiword examples.
-    if " " in word:
+    if (" " in word or "\u00A0" in word) and not no_skip_spaces:
```
kylebgorman: Ditto.

kylebgorman:
I just added Jackson in case he's available.

lfashby:
> One thing that makes this generally hard to follow is that the flag semantics have a negation in them. Maybe it would be easier to call it something such that it's a True and not a False when it's enabled...any ideas?

I agree. I confused myself many times while adding the flags. Perhaps more sensible flag names would be `allow_spaces_words`/`allow_spaces_prons`. I suppose I was influenced by the amount of `--no-...` type flags that we already have. 

A question I have is --
Jackson's [comment](https://github.com/kylebgorman/wikipron/issues/91#issuecomment-552208879) from #91 : 
> In general, I think our API design should be "if the user wants to remove information, they should be forced to do something, e.g., explicitly applying a flag to say so". 

Seems to suggest (if 'information' can mean number of transcriptions we scrape) that the default behavior should _not_ be to skip entries with spaces and that using the flag should compel WikiPron to skip entries with spaces. As the flags are currently implemented, our default behavior is to skip entries with spaces, but I wonder if it would be more consistent with Jackson's comments to have the reverse be the case.

kylebgorman:
Lot of good points here.

I think we should accept this PR as is (modulo my point about “word” vs.
“ortho”) and then at a later date, take Jackson’s advice and refactor for
that.

What do you think?

yeonju123:
I am working on it too, but mine is not complete yet. 

Lucas’s looks great :D

Yeonju

languages/wikipron/src/scrape.py
lfashby: Let me know if you think my solution for this is alright. It uses the least amount of lookups I could come up with, but does entail looping through all of each sub-dictionary for every languages.

lfashby:
I made the requested changes. Should I have totally changed the flag names to something like `include_spaces_word` or will will leave them as they are (`no_skip_spaces_word`/`no_skip_spaces_pron`) for the time being? 
Yeonju's solution #137 of using a single flag for both is pretty cool (I'd imagine people would probably use both of my provided flags or neither) and her command line help statements are better so perhaps I should do some further changes to incorporate her ideas?

wikipron/cli.py
```diff
@@ -87,5 +97,6 @@ def _scrape_and_write(config: Config) -> None:
 def main() -> None:
     logging.basicConfig(format="%(levelname)s: %(message)s", level="INFO")
     args = _get_cli_args(sys.argv[1:])
+    print(args)
```
kylebgorman: Still want to remove this?

kylebgorman:
I agree re: avoiding (double)-negation in flags; please go for it.

Welcome back Jackson! Hope you're well.

lfashby:
Had to resolve conflicts (in `CHANGELOG.md` and `src/scrape.py`) in order to merge.

PR #136, opened 2020-03-29
add no_skip_space flag

yeonju123:
`--no-skip-space` when used will not skip words that contain a space. This required editing `config.py`, `scrape.py`, `cli.py`, and `extract/core.py`.

(I understand that Lucas already created PR, but I am creating this PR wondering if some portion of my code can be useful. If possible, feedback will be appreciated!)


wikipron/scrape.py
```diff
@@ -14,16 +14,19 @@
...
-        return True
+def _skip_word(word: str, no_skip_space: bool) -> bool:
     # Skips examples containing a dash.
     if "-" in word:
         return True
     # Skips examples containing digits.
     if re.search(r"\d", word):
         return True
+    if no_skip_space:
+        if " " in word:
```
kylebgorman: you don't need this inner check: it can simply be:

`if no_skip_space: return False`.

Or better yet you'd replace the entire block 24-29 with something like:

```
if not no_skip_space and " " in word: return True
else: return False
```

PR #139, opened 2020-04-16
Two more Vietnamese dialects are added.

yeonju123:

Two dialects of Vietnamese, Thanh Chương and Hà Tĩnh, are added in `languages.json`.


lfashby:
This looks good to me as far as the Vietnamese dialects we should add and I think we should merge this in. Apparently Vinh, Thanh Chương and Hà Tĩnh are all in the North Central dialect region. In [this example](https://en.wiktionary.org/wiki/c%E1%BA%A7m) they are all listed together, but [here](https://en.wiktionary.org/wiki/ch%C3%A1y) Hà Tĩnh gets it's own transcription. 
I'm going to do another scrape with a modified version of this dialect objects to see if Vinh and Thanh Chương on their own scrape any entries and how many entries we get if Vinh, Thanh Chương and Hà Tĩnh are all grouped together, but for the time being I think the dialect object in this PR is the one we should go with.

yeonju123:
I think that’s a good idea since we don’t know which dialect combinations are listed together. 

There may be Thanh Chương and Hà Tĩnh combination as well.

kylebgorman:
Does this still need reviewed? I guess this just needs to be added to the changelog and then it's ready to go in.

PR #145, opened 2020-04-17
Add commas to the counts in the README table

kylebgorman:
`generate_summary.py` is lightly modified; the `README.md` table is
regenerated.


lfashby:
This looks mostly fine but I have a few comments.

languages/wikipron/src/generate_summary.py
```diff
@@ -21,6 +21,8 @@ def _handle_wiki_name(
     for modifier in modifiers:
         if modifier in language:
             key = file_path[file_path.index("_") + 1:file_path.rindex("_")]
+            if not key:
+                continue
             values = language[modifier][key]
```
lfashby, line 26: Did a `KeyError` with `values = languages[modifier][key]` prompt you to add this?  An error like that is something we need to know about so I think logging `file_path` or `language[wiktionary_name]` in `if not key:` would be good.

We can also prepare for Aaron adding another `_` to our TSV names with a slight modification like the following:
```
        if modifier in language:
            k_start = file_path.index("_") + 1
            key = file_path[k_start:file_path.index("_", k_start)]
            ...
```

languages/wikipron/src/generate_summary.py
```diff
@@ -74,38 +76,28 @@ def main() -> None:
...
+        )
+        print(
+            "| :---- | :----: | :----: | :----: | :----: | :----: | ----: |",
+            file=sink,
+        )
+        for link, code, n1, n2, cf, ph, count in readme_list:
+            print(
+                f"| {link} | {code} | {n1} | {n2} | {cf} | {ph} | {count:,} |",
+                file=sink,
+            )
```
lfashby, line 100: This is certainly a cleaner way of doing it, my only complaint would be `n1, n2, cf, ph` are not particularly friendly/informative variable names.

languages/wikipron/src/generate_summary.py
kylebgorman: Yeah, what should the log message say?

languages/wikipron/src/generate_summary.py
kylebgorman: Sure, but I think this a rare case where it doesn't matter and giving them informative ones would blow up the line-wrapping so badly that the net cognitive effect would be negative.

languages/wikipron/src/generate_summary.py
lfashby: Something like:
```
            logging.info(
                'Failed to isolate key for "%s" modifier in "%s".',
                modifier,
                file_path,
            )
```
Would give us - Failed to isolate key for "script" modifier in "hbs_cyrl_phonemic.tsv".
That ought to be enough information for us to get to the bottom of whatever the problem might be. Did you encounter a `KeyError` after running a scrape of Latin with the dialects? If so that is because you have to get rid of the old Latin files (`lat_phonemic.tsv` and `lat_phonetic.tsv`) before running `generate_summary.py`. Regardless, it's a good idea to catch this error.

kylebgorman: Done.

Issue #143, opened 2020-04-17
[lat] Dialect specifications


kylebgorman:
Add "Classical", "Ecclesiastical", and "Vulgar" specifications for Latin.

kylebgorman:
Not directly related but the Latin extractor is also picking up pronunciations from other languages, like [this Spanish one](https://en.wiktionary.org/wiki/grammatica#Pronunciation_3).

lfashby:
> Not directly related but the Latin extractor is also picking up pronunciations from other languages, like [this Spanish one](https://en.wiktionary.org/wiki/grammatica#Pronunciation_3).

That's quite an interesting problem. It looks to me like the root of the problem is in how the pron selector is defined, in particular how `following-sibling` as defined [here](https://github.com/kylebgorman/wikipron/blob/master/wikipron/extract/lat.py#L67) works. 
The first following-sibling `<ul>` of:
```
<h3>
  <span class="mw-headline" id="Etymology_2">Etymology 2</span>
  ...
</h3>
```
(which is the element that satisfies: `//{heading}[span[@class = "mw-headline" and @id = "{tag}"]]`) is the `<ul>` containing the Spanish pronunciation.

Here is a [snippet of code](https://gist.github.com/lfashby/e940644e31f8caa8f0fc18d7e7666236) I use to test the extraction functions, it is crude but allows you to test how an extraction function works on a particular entry (specified by providing the Wiktionary URL for that entry).

kylebgorman:
Thanks for that. I'm really bad at reasoning about these so perhaps I can
convince somebody else to tackle this.

PR #146, opened 2020-04-18
Latin dialects

kylebgorman:
This is not working: the dialect files are identical; but am just sharing in case I have missed something.


lfashby:
Do you mean that the you scraped Latin after adding these dialects to `languages.json` and the contents of the TSVs for the different dialects are all the same?

kylebgorman:
Sorry, this is a terribly non-descript PR!

Yes, when I run with a (truncated for speed) `languages.json` the Ecclesiastical and Classical files are unexpectedly identical and both contain the prons labeled "Ecclesiastical" and "Classical", respectively. [Sample word with both of those dialects here](https://en.wiktionary.org/wiki/eruptio#Latin).

kylebgorman:
Also a bunch of unrelated changes have ended up in this PR...I have too many PRs at once. But the relevant one is to `languages.json`.

lfashby:
I think when we added the Latin extraction function (and our other extraction functions that build their own pron selectors and don't rely on `_PRON_XPATH_SELECTOR_TEMPLATE` from `config.py`) we did not consider adding dialect support.

kylebgorman:
Okay, so if there's a language customization in `extract`, what exactly happens to the dialect specifications? Are they completely ignored, or have to be added to the language customization, or something else?

lfashby:
As far as I can tell, languages that do not interact with `extract_word_pron_default` (and in particular `_yield_phn` and its use of `config.pron_xpath_selector` ) in `extract/default.py` do not have any dialect support - meaning all languages for which we have extraction functions just ignore dialects.

kylebgorman:
At the very least we should document (or log loudly?) this.

And while it's not super important not sure how useful the Latin data is
without dialect specifications.

I feel like I'm out of my depth with figuring out how that works given how
elaborate the Latin extraction function is.

lfashby:
I created an issue to track this.

jacksonllee:
Lucas is correct that languages which require a non-default extraction treatment ignore the dialect labels unless they are specifically used in the respective extraction function. We should figure out how the extraction module for Latin can make use of `config.pron_xpath_selector` somehow (as the Japanese one does, as Lucas has pointed out). If I'm not mistaken, it would be a matter of combining the Latin-specific `pron_xpath_selector` [here](https://github.com/kylebgorman/wikipron/blob/2b36c9c3f971bb390c197d2581aa536abb751282/wikipron/extract/lat.py#L114-L115) with `config.pron_xpath_selector` that has the dialect info.

kylebgorman:
This is quite out of date but ongoing work is happening on #143.

PR #151, opened 2020-04-27
Enguk

mam0chan:


kylebgorman:
File looks good! Could you also sort the file (`sort -o nameoffile nameofile` and then add, commit, push) before we pull?

kylebgorman:
The way the whitelist format will work (and my Latin one worked) we can put comments into the file using `#` as a comment delimiter. It seems like a some of the rarer ones here are probably worthy of comment. I would have thought that /ɫ/ should be just an allophone, and that glottal stop should not be listed at all. But if we want to leave them in we can just leave notes to ourselves to go and investigate later. Also there are surely too many vowel phonemes but I'm not sure which are the most egregious.

kylebgorman:
Copying comments from offline discussion for reference:

- You also have /l/ in there on line 6 or 7; maybe just annotate the velar l with a comment # Allophone of /l/. or something like that.
- Glottal stop is marginal as a phoneme. When it's word-initial you can think of it as an allophone ("of the left edge of a word") that appears in vowel-initial words. The only word I can think of that has a glottal stop in any other position is "uh-oh". If you want to keep it I would just say # Marginal..
- For /r/ can you add a comment # Mistake for /ɹ/. or something like that? When we get around to writing a "bot" to correct this we can fix it en-masse and remove it from the list
- I don't have any strong opinion about how the <ch> and <j> affricates should be done, whether they're "contour phonemes" or two phonemes, but in the data they're usually transcribed with a tie-bar so we can leave it as is, seems fine. I've never seen <let's> transcribed with a tiebar, that seems totally unnecessary both from a phonemic and a phonetic perspective.

PR #152, opened 2020-05-01
Adds whitelist.py

AaronGoyzueta:
Creates a script that, when called on a phonemic TSV in the TSV folder, finds the appropriate whitelist and creates a new TSV with only characters found in the whitelist.

### DIFF
--- a/languages/wikipron/src/postprocess
+++ b/languages/wikipron/src/postprocess
@@ -6,8 +6,7 @@ for TSV in ../tsv/*.tsv; do
     # rather than a locale-dependent comparison.
     LC_ALL=C sort -u -o "${TSV}" "${TSV}"
     ./generalized_split.py "${TSV}"
-
 done
-./whitelist.py "../tsv/eng_uk_phonemic.tsv"
-./whitelist.py "../tsv/eng_us_phonemic.tsv"
-./whitelist.py "../tsv/lat_phonemic.tsv"
+./whitelist.py ../tsv/eng_uk_phonemic.tsv ../whitelist/eng_uk_phonemic.whitelist ../tsv/eng_uk_phonemic.tsv
+./whitelist.py ../tsv/eng_us_phonemic.tsv ../whitelist/eng_us_phonemic.whitelist ../tsv/eng_us_phonemic.tsv
+./whitelist.py ../tsv/lat_phonemic.tsv ../whitelist/lat_phonemic.whitelist ../tsv/lat_filtered_phonemic.tsv
--- a/languages/wikipron/src/postprocess
+++ b/languages/wikipron/src/postprocess
@@ -1,0 +1,13 @@
+#!/bin/bash
+set -eou pipefail
+
+for TSV in ../tsv/*.tsv; do
+    # Explicitly uses byte-wise comparison for sorting
+    # rather than a locale-dependent comparison.
+    LC_ALL=C sort -u -o "${TSV}" "${TSV}"
+    ./generalized_split.py "${TSV}"
+
+done
+./whitelist.py "../tsv/eng_uk_phonemic.tsv"
+./whitelist.py "../tsv/eng_us_phonemic.tsv"
+./whitelist.py "../tsv/lat_phonemic.tsv"
--- a/languages/wikipron/src/postprocess
+++ b/languages/wikipron/src/postprocess
@@ -7,6 +7,7 @@ for TSV in ../tsv/*.tsv; do
     LC_ALL=C sort -u -o "${TSV}" "${TSV}"
     ./generalized_split.py "${TSV}"
 done
+
 ./whitelist.py ../tsv/eng_uk_phonemic.tsv ../whitelist/eng_uk_phonemic.whitelist ../tsv/eng_uk_phonemic.tsv
 ./whitelist.py ../tsv/eng_us_phonemic.tsv ../whitelist/eng_us_phonemic.whitelist ../tsv/eng_us_phonemic.tsv
 ./whitelist.py ../tsv/lat_phonemic.tsv ../whitelist/lat_phonemic.whitelist ../tsv/lat_filtered_phonemic.tsv
--- a/languages/wikipron/src/whitelist.py
+++ b/languages/wikipron/src/whitelist.py
@@ -1,45 +1,43 @@
 #!/usr/bin/env python
+"""Filters TSVs through a whitelist file.
 
-import sys
+This script takes lines from a source TSV and writes only lines with
+pronunciations containing characters found in the whitelist file to a new TSV."""
+
+import argparse
 import re
+import sys
+
+from typing import FrozenSet, Iterator
 
 
-def whitelist_reader(path):
-    """Reads in whitelist file"""
+def whitelist_reader(path: str) -> Iterator[str]:
+    """Reads in whitelist file."""
     with open(path, "r") as source:
         for line in source:
-            yield line.rstrip().split()[0]
+            # Removes comments from line.
+            line = re.sub("\s.*", "", line)
+            yield line
 
 
-def filter(tsv_path, phonemes, output_path):
-    """Creates tsv filtered by whitelist"""
-    with open(tsv_path, "r") as source:
-        with open(output_path, "w") as output:
-            for line in source:
-                pron = set(line.split("\t")[1].split())
-                if phonemes.issuperset(pron):
-                    print(line.rstrip(), file=output)
-                else:
-                    pass
+def filter_and_write(tsv_path: str, phones: FrozenSet[str], output_path: str) -> None:
+    """Creates TSV filtered by whitelist."""
+    with open(tsv_path, "r") as source, open(output_path, "w") as output:
+        for line in source:
+            pron = line.split("\t")[1]
+            these_prons = frozenset(pron.split())
+            if phones.issuperset(these_prons):
+                print(line, file=output, end='')
 
 
-def main() -> None:
-    tsv_path = sys.argv[1]
-    iso639_code = tsv_path[tsv_path.rindex("/") + 1:tsv_path.index("_")]
-    dialect = re.search("_(.*)_", tsv_path)
-    if dialect:
-        whitelist_path = f"../whitelist/{iso639_code}_{dialect[1]}_phonemic.whitelist"
-        output_path = f"../tsv/{iso639_code}_{dialect[1]}_filtered_phonemic.tsv"
-    else:
-        whitelist_path = f"../whitelist/{iso639_code}_phonemic.whitelist"
-        output_path = f"../tsv/{iso639_code}_filtered_phonemic.tsv"
-    try:
-        whitelist_phonemes = set(whitelist_reader(whitelist_path))
-        whitelist_phonemes.add(" ")
-        filter(tsv_path, whitelist_phonemes, output_path)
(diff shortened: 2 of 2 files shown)


languages/wikipron/src/postprocess
```diff
@@ -0,0 +1,13 @@
...
+set -eou pipefail
+
+for TSV in ../tsv/*.tsv; do
+    # Explicitly uses byte-wise comparison for sorting
+    # rather than a locale-dependent comparison.
+    LC_ALL=C sort -u -o "${TSV}" "${TSV}"
+    ./generalized_split.py "${TSV}"
+
+done
+./whitelist.py "../tsv/eng_uk_phonemic.tsv"
```
kylebgorman, line 11: Looks good. A few small requests:


* Could you move the blank line on line 9 to after `done` so that it reads roughly

```
...
     ./generalized_split.py "${TSV}"
done

./whitelist.py ../tsv/eng_uk_phonemic.tsv
...
```

* You don't need to quote file names if they don't include spaces, wild-card expansions, or variable expansions, as is the case here.

languages/wikipron/src/whitelist.py
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env python
+
```
kylebgorman, line 2: Put a documentation string here explaining what this does in a single sentence, with a period at the end,

languages/wikipron/src/whitelist.py
```diff
@@ -0,0 +1,45 @@
...
+import sys
```
kylebgorman, line 3: Sort imports lexicographically; e.g., import `re` before `sys`.

FYI, the best-practice rules for imports are:

* imports from the standard library (except `typing`), in lexicographic order, using the `import X` style
* a blank line
* imports from `typing`, using the `from typing import X` style
* a blank line
* imports from third-party libraries (e.g., installable from pip), in lexicographic order, using the `import X` style
* a blank line
* any module-specific imports, in lexicographic order, using the `import X` style

(You can omit the blank lines between blocks if any of those blocks are empty.)

languages/wikipron/src/whitelist.py
```diff
@@ -0,0 +1,45 @@
...
+
+def whitelist_reader(path):
+    """Reads in whitelist file"""
+    with open(path, "r") as source:
+        for line in source:
+            yield line.rstrip().split()[0]
+
+
+def filter(tsv_path, phonemes, output_path):
+    """Creates tsv filtered by whitelist"""
```
kylebgorman, line 15: `tsv` should be `TSV` in a prose comment: it's an acronym. Add a period at the end of this doc-string.

languages/wikipron/src/whitelist.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def whitelist_reader(path):
+    """Reads in whitelist file"""
+    with open(path, "r") as source:
+        for line in source:
+            yield line.rstrip().split()[0]
+
+
+def filter(tsv_path, phonemes, output_path):
```
kylebgorman, line 14: The function `filter` suggests that it's going to return something but instead it writes out the filtered file too! I would call it something like `filter_and_write` or `filter_write`.

Can you type this function? `tsv_path` and `output_path` are `str`, `phonemes` (should be renamed `phones`, since this should also work on "phonetic" data and that's a generic term that encompasses both) should have type `FrozenSet[str]` (have to import it). And the return type is `None`. Then you can check the typing using `mypy` (install it from `pip` then run `mypy whitelist.py`.)

languages/wikipron/src/whitelist.py
```diff
@@ -0,0 +1,45 @@
...
+    """Creates tsv filtered by whitelist"""
+    with open(tsv_path, "r") as source:
+        with open(output_path, "w") as output:
+            for line in source:
+                pron = set(line.split("\t")[1].split())
```
kylebgorman, line 19: I would break this apart into two pieces to make it easier to read.

    pron = line.split("\t")[1]
    these_phones = frozenset(pron.split())

I'm also using `frozenset` here because you're not mutating the set.

languages/wikipron/src/whitelist.py
```diff
@@ -0,0 +1,45 @@
...
+                if phonemes.issuperset(pron):
+                    print(line.rstrip(), file=output)
+                else:
```
kylebgorman, line 22: Omit the `else` and the pass. It does no work.

languages/wikipron/src/whitelist.py
```diff
@@ -0,0 +1,45 @@
...
+                    pass
+
+
+def main() -> None:
+    tsv_path = sys.argv[1]
```
kylebgorman, line 27: This will crash unceremoniously if someone doesn't provide a file as the command-line argument. Use `argparse` instead, like we did for the `evaluate_all.py` script you wrote and put the argument parsing inside `if __name__ == __main__:`, passing it to `main`.

languages/wikipron/src/whitelist.py
```diff
@@ -0,0 +1,45 @@
...
+    iso639_code = tsv_path[tsv_path.rindex("/") + 1:tsv_path.index("_")]
+    dialect = re.search("_(.*)_", tsv_path)
+    if dialect:
+        whitelist_path = f"../whitelist/{iso639_code}_{dialect[1]}_phonemic.whitelist"
+        output_path = f"../tsv/{iso639_code}_{dialect[1]}_filtered_phonemic.tsv"
+    else:
+        whitelist_path = f"../whitelist/{iso639_code}_phonemic.whitelist"
+        output_path = f"../tsv/{iso639_code}_filtered_phonemic.tsv"
+    try:
```
kylebgorman, line 36: I don't think this try/except behavior is wise. We would like to know if a file isn't found, for whatever reason. So I would simply eliminate the `try`/`except` bit so we get a sensible error.

languages/wikipron/src/whitelist.py
```diff
@@ -0,0 +1,45 @@
...
+                pron = set(line.split("\t")[1].split())
+                if phonemes.issuperset(pron):
+                    print(line.rstrip(), file=output)
+                else:
+                    pass
+
+
+def main() -> None:
+    tsv_path = sys.argv[1]
+    iso639_code = tsv_path[tsv_path.rindex("/") + 1:tsv_path.index("_")]
```
kylebgorman, line 28: Can you add some comments about what is happening here? This is complex code.

I am also not sure we should have special handling for dialect; or rather, if there's a dialect specification, we should assume that the whitelist files are also dialect-specific. That should simplify the logic here quite a bit. And this seems to be hard-coded to use phonemic files, which is not intended.

Let me suggest a simpler setup: just have the user specify all three files as command-line arguments, at least for now. We can figure something else out when/if @lfashby incorporates this into `languages.json`.

languages/wikipron/src/whitelist.py
```diff
@@ -0,0 +1,45 @@
...
+    dialect = re.search("_(.*)_", tsv_path)
+    if dialect:
+        whitelist_path = f"../whitelist/{iso639_code}_{dialect[1]}_phonemic.whitelist"
+        output_path = f"../tsv/{iso639_code}_{dialect[1]}_filtered_phonemic.tsv"
+    else:
+        whitelist_path = f"../whitelist/{iso639_code}_phonemic.whitelist"
+        output_path = f"../tsv/{iso639_code}_filtered_phonemic.tsv"
+    try:
+        whitelist_phonemes = set(whitelist_reader(whitelist_path))
```
kylebgorman, line 37: Use `frozenset` here instead.

languages/wikipron/src/whitelist.py
```diff
@@ -0,0 +1,45 @@
...
+def whitelist_reader(path):
+    """Reads in whitelist file"""
+    with open(path, "r") as source:
+        for line in source:
+            yield line.rstrip().split()[0]
+
+
+def filter(tsv_path, phonemes, output_path):
+    """Creates tsv filtered by whitelist"""
+    with open(tsv_path, "r") as source:
```
kylebgorman, line 16: You can put two `with` statements on a single line, if under 80 lines. This will look something like:

`with open(tsv_path, "r") as source, open(output_path, "w") as output:`

Then adjust the indentation below.

languages/wikipron/src/whitelist.py
```diff
@@ -0,0 +1,45 @@
...
+        with open(output_path, "w") as output:
+            for line in source:
+                pron = set(line.split("\t")[1].split())
+                if phonemes.issuperset(pron):
+                    print(line.rstrip(), file=output)
```
kylebgorman, line 21: You can also do: `print(line, file=output, end='')`, which prevents it from adding a newline.

languages/wikipron/src/whitelist.py
```diff
@@ -0,0 +1,45 @@
...
+
+import sys
+import re
+
+
+def whitelist_reader(path):
+    """Reads in whitelist file"""
+    with open(path, "r") as source:
+        for line in source:
+            yield line.rstrip().split()[0]
```
kylebgorman, line 11: Please type this function. It should be:

   `def whitelist_reader(path: str) -> Iterator[str]:`

Is this splitting to separate off the comment character? Feels a bit non-robust. I would instead suggest using a regular expression substitution (`re.sub`) to remove the comments.

languages/wikipron/src/whitelist.py
```diff
@@ -0,0 +1,45 @@
...
+    dialect = re.search("_(.*)_", tsv_path)
+    if dialect:
+        whitelist_path = f"../whitelist/{iso639_code}_{dialect[1]}_phonemic.whitelist"
+        output_path = f"../tsv/{iso639_code}_{dialect[1]}_filtered_phonemic.tsv"
+    else:
+        whitelist_path = f"../whitelist/{iso639_code}_phonemic.whitelist"
+        output_path = f"../tsv/{iso639_code}_filtered_phonemic.tsv"
+    try:
+        whitelist_phonemes = set(whitelist_reader(whitelist_path))
+        whitelist_phonemes.add(" ")
```
kylebgorman, line 38: This should not be necessary, and looks hackish.

languages/wikipron/src/whitelist.py
```diff
@@ -1,45 +1,43 @@
...
-        whitelist_path = f"../whitelist/{iso639_code}_phonemic.whitelist"
-        output_path = f"../tsv/{iso639_code}_filtered_phonemic.tsv"
-    try:
-        whitelist_phonemes = set(whitelist_reader(whitelist_path))
-        whitelist_phonemes.add(" ")
-        filter(tsv_path, whitelist_phonemes, output_path)
-    except FileNotFoundError:
-        pass
+            # Removes comments from line.
+            line = re.sub("\s.*", "", line)
```
kylebgorman, line 19: This will just delete everything after space. But that's not exactly how a comment works. A comment is defined as optional space `\s*` mandatorily followed by `#` then the rest of the line `.*$`; I think that would be a bit more constrained.

AaronGoyzueta: Okay, that makes sense, should I push this change now?

kylebgorman: Please make that change, test, and then add, commit, and push the change. Then you can press the big green "Squash and merge" button" (I think you should have the privileges to do that), or you can ping me and I'll do it.

Issue #153, opened 2020-05-02
Single-language "big scrape" functionality


kylebgorman:
Often we want to test changes to one or a small number of languages by running a partial "big scrape" and it would be nice to make this easier to do than making copies of `languages.json`. Maybe something like:

    ./scrape.py --restriction=lit

Would run the scrape for Lithuanian (all dialects and scripts) but would use the appropriate metadata in `languages.json`, and 

    ./scrape.py

would run it for **all** languages in `languages.json`.

bonham79:
Just sent pull with --restriction flag option for one or more languages.

PR #154, opened 2020-05-03
Added --restriction flag

bonham79:
Added --restriction flag to limit 'big scrape' to individual languages. Can specify one or more languages to scrape by following flag with ISO codes of desired languages. 

### DIFF
--- a/data/wikipron/src/scrape.py
+++ b/data/wikipron/src/scrape.py
@@ -66,7 +66,7 @@ def main(args: argparse.Namespace) -> None:
 
     codes = list(languages.keys())
 
-    # Verifies mypy language code for --restriction is valid
+    # Verifies language code for --restriction is valid
     if args.restriction:
         # Cleans entry.
         keys = re.split(r"[;,\s]+\s*", args.restriction[0].strip(";, "))
@@ -130,7 +130,6 @@ if __name__ == "__main__":
         level="INFO",
     )
 
-    # Check for --restriction flag
     parser = argparse.ArgumentParser()
     parser.add_argument(
         "--restriction",
--- a/data/wikipron/src/scrape.py
+++ b/data/wikipron/src/scrape.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 """Performs the big scrape."""
 
-
+import argparse
 import datetime
 import json
 import logging
@@ -59,9 +59,20 @@ def _build_scraping_config(
     _call_scrape(config_settings, phonetic_config, phonetic_path)
 
 
-def main() -> None:
+def main(args) -> None:
     with open(LANGUAGES_PATH, "r") as source:
         languages = json.load(source)
+
+    ###Checks restriction codes are valid.
+    if args.restriction: 
+        for key in args.restriction:
+            try:
+                assert key in languages 
+            except AssertionError:
+                print("Language restrictions must be valid ISO code.")
+                exit()
+    ###
+
     # "2020-01-15" (Big Scrape 3).
     cut_off_date = datetime.date.today().isoformat()
     wikipron_accepted_settings = {
@@ -71,6 +82,12 @@ def main() -> None:
     }
 
     for iso639_code in languages:
+
+        ###Checks for language restrictions
+        if args.restriction and iso639_code not in args.restriction:
+            continue
+        ###
+
         language_settings = languages[iso639_code]
         for k, v in language_settings.items():
             if k in wikipron_accepted_settings:
@@ -108,4 +125,11 @@ if __name__ == "__main__":
         datefmt="%d-%b-%y %H:%M:%S",
         level="INFO",
     )
-    main()
+
+    ###Add parser to limit scope of scrape. 
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--restriction', type=str, nargs= '*', default=False, help='Specify language restrictions for scrape')
+    args = parser.parse_args()
+    ###   
+    
+    main(args)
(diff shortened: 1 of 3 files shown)


kylebgorman:
Cool! I'm going to add @lfashby to review; looks basically right but maybe some style issues.

Could you type `args` when it's passed to `main`? E.g.:

    def main(args: argparse.Namespace) -> None:

Could you add "Closes #153" to the PR itself? That'll cause GitHub to automatically close the issue when/if this is merged.

kylebgorman:
Also do add this to the CHANGELOG as detailed above, and check the check-box when you do.

kylebgorman:
Probably should add a line to `data/wikipron/src/README.md` to indicate this feature is available and you don't need to make a copy of `languages.json` to run a partial scrape.

kylebgorman:
* Can you use a standard "# foo" comment style (space after `#`, just one per line, declarative sentence etc.)?
* If you don't specify `False` as the default for `--restriction` you'll just get `None` as the default, which seems desirable. I think the relevant conditionals should still work, though.

lfashby:
I think this looks alright. There is only one thing I would add in addition to Kyle's comments.

For the record, I'm not convinced that the big scrape should have this sort of functionality. I guess if it makes our lives easier (with respect to testing how WikiPron works on a subset of languages) then it is fine to have, but it doesn't address the currently awkward part of running a big scrape -- having to manually find the languages WikiPron failed on during the big scrape and then restricting `languages.json` to just those languages. In the future we could automate the the process of building a `languages.json`-esque object for languages that WikiPron failed on during the big scrape so we don't have to manually modify `languages.json` to retry failed languages.

data/wikipron/src/scrape.py
```diff
@@ -59,9 +59,20 @@ def _build_scraping_config(
...
     with open(LANGUAGES_PATH, "r") as source:
         languages = json.load(source)
+
+    ###Checks restriction codes are valid.
+    if args.restriction: 
+        for key in args.restriction:
+            try:
+                assert key in languages 
+            except AssertionError:
+                print("Language restrictions must be valid ISO code.")
```
lfashby, line 72: Could you change this to `logging.info()` and log the failed `key`?

kylebgorman: This should also be a fatal error. While `exit()` does kill things it just makes it look like it succeeded because the return code of the program will be 0 (non-zero means "failure"). I would use `logging.fatal` instead and then `exit(1)` or something like that.

data/wikipron/src/scrape.py
```diff
@@ -71,6 +82,12 @@ def main() -> None:
     }
 
     for iso639_code in languages:
+
+        ###Checks for language restrictions
+        if args.restriction and iso639_code not in args.restriction:
```
kylebgorman, line 87: Here's an alternative that's maybe more complicated but more efficient. Having written it out I'm not sure it's an improvement, so take this with a grain of salt.

    codes = list(languages.keys())
    if args.restriction:
       rset = frozenset(args.restriction)
       lset = frozenset(codes)
       if rset - lset: 
          # fatal error here: log the missing keys and exit(1)
      codes = list(rset & lset)

Then the loop is just:

     for code in codes:
         language_settings = languages[code]  
         ...

data/wikipron/src/scrape.py
```diff
@@ -58,10 +59,29 @@ def _build_scraping_config(
...
     with open(LANGUAGES_PATH, "r") as source:
         languages = json.load(source)
+
+    codes = list(languages.keys())
+
+    # Verifies language code for --restriction is valid
+    if args.restriction:
+        # Cleans entry. 
+        keys = re.split(r'[;,\s]+\s*',args.restriction[0].strip(';, '))
+        if not keys[0]:
```
kylebgorman, line 72: I think you can say `nargs="+"` to require at least one and make this logic unnecessary?

data/wikipron/src/scrape.py
bonham79: Definitely better. Throwing it in. (Though isn't searching rather cheap here?)

data/wikipron/src/scrape.py
kylebgorman: Actually, sorry to keep changing but I think also you can do something like:

    --restriction=fre --restriction=spa

to have multiple values. Then you don't need to do the `re`-splitting (and will no longer need `re`).

data/wikipron/src/scrape.py
```diff
@@ -108,4 +128,10 @@ def main() -> None:
         datefmt="%d-%b-%y %H:%M:%S",
         level="INFO",
     )
-    main()
+
+    # Check for --restriction flag
```
kylebgorman, line 133: Comment not current anymore?

data/wikipron/src/scrape.py
bonham79: The argparse is registering empty string cases even with nargs="+." So if --restriction= is passed then it will search for '' as a string and then log '' is not a valid ISO code. Which doesn't seem informative for an error.

kylebgorman:
Two nits left then we're ready. I am looking forward to having this around.

To reflow on this project we use `black -l79 scrape.py`, then check `flake8 scrape.py` (for linting) and `mypy scrape.py` (for static typing errors). All three---`black`, `flake8`, and `mypy`---are on PyPi).

data/wikipron/src/scrape.py
kylebgorman: Ah, didn't realize that. Not very useful, I agree, keep this then.

data/wikipron/src/scrape.py
bonham79: Let me get that real quick.

Issue #155, opened 2020-05-03
Incorporate whitelisting into `languages.json`


kylebgorman:
Once #152 is merged we should merge the whitelist functionality into `languages.json`. Presumably it should simply be the case that for any language or dialect if a `whitelist` value is specified, we load it and generate open two TSV files, one for the total data, one for the filtered data. This would also in theory make `whitelist.py` obsolete.

I would suggest that we actually call the unfiltered one a special name or put it somewhere different.

bonham79:
Any particular reason for making it part of languages.json instead of just searching for a whitelist file and defaulting to unfiltered if not found? The search isn't particularly costly and troubleshooting could be limited to checking the whitelist file instead of that and languages.json.

kylebgorman:
That would be fine too I suppose, and an improvement from what we currently have. It's a question of "automagic" vs. explicit configuration.

If you go that route, I would just make sure the log captures useful information from the whitelisting---e.g., that a whitelist was found, how many things were filtered, etc.

PR #157, opened 2020-05-03
Whitelist cleanup

### LINKED ISSUE #156: Add logging to whitelister
Once #152 is in we should add some kind of logging to the whitelister so we can get a list of "bad" words and the phones that go with them.

Perhaps it should also produce a some kind of structured report (JSON?) instead of just logging.

kylebgorman:
* Logs bad phonemes (closes #156)
* Makes whitelist functions semiprivate
* Factors `postprocess` script into `split` and `whitelist` portions; makes it easier to run one but not the other
* Makes whitelisting sensitive to the presence of whitelists: if one exist, it is applied
* Runs this process over Latin and regenerates the summaries

FYI: @AaronGoyzueta @jacksonllee

### DIFF
--- a/data/wikipron/src/whitelist.py
+++ b/data/wikipron/src/whitelist.py
@@ -1,43 +1,55 @@
 #!/usr/bin/env python
 """Filters TSVs through a whitelist file.
 
-This script takes lines from a source TSV and writes only lines with
-pronunciations containing characters found in the whitelist file to a new TSV."""
+This script takes lines from a source TSV and writes only entries whose
+pronunciations only contain phones found on the whitelist into a new TSV file.
+"""
 
 import argparse
+import logging
 import re
-import sys
 
 from typing import FrozenSet, Iterator
 
 
-def whitelist_reader(path: str) -> Iterator[str]:
+def _whitelist_reader(path: str) -> Iterator[str]:
     """Reads in whitelist file."""
     with open(path, "r") as source:
         for line in source:
-            # Removes comments from line.
-            line = re.sub("\s*#.*$", "", line)
+            line = re.sub(r"\s*#.*$", "", line)  # Removes comments from line.
             yield line.rstrip()
 
 
-def filter_and_write(tsv_path: str, phones: FrozenSet[str], output_path: str) -> None:
+def _filter_write(
+    input_tsv_path: str, phones: FrozenSet[str], output_tsv_path: str
+) -> None:
     """Creates TSV filtered by whitelist."""
-    with open(tsv_path, "r") as source, open(output_path, "w") as output:
+    with open(input_tsv_path, "r") as source, open(
+        output_tsv_path, "w"
+    ) as sink:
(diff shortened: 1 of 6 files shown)


lfashby:
This looks good, just a few comments.

Regarding logging, I like the idea of a 'structured report' mentioned in #156, but I think this ought to be fine. Perhaps if we direct the logging statements from `whitelist.py` to their own file (as is done with the logging statements generated in `src/scrape.py`) they will be easier to examine/work with? 

Currently this is going to overwrite our old TSV's, which for the time being seems alright and `INPUT_TSV="${WHITELIST//whitelist/tsv}"` is an elegant solution. Eventually maybe we want to keep both the filtered and unfiltered data and give the filtered data a different name as mentioned in #155.

data/wikipron/src/whitelist.py
```diff
@@ -1,43 +1,55 @@
...
 
 def main(args: argparse.Namespace) -> None:
-    whitelist_phones = frozenset(whitelist_reader(args.whitelist_path))
-    filter_and_write(args.tsv_path, whitelist_phones, args.output_path)
+    whitelist_phones = frozenset(_whitelist_reader(args.whitelist_path))
+    _filter_write(args.input_tsv_path, whitelist_phones, args.output_tsv_path)
 
 
 if __name__ == "__main__":
+    logging.basicConfig(
```
lfashby, line 48: I tested this out locally and found that running `whitelist.py` from the command line will generate the desired logging statements, but running it as a part of `postprocess` will not. I guess the `sys.stderr` that `whitelist.py` is trying to write to when it is run within the bash script is not visible? 
Adding a handler like 
```
        handlers=[
            logging.StreamHandler(sys.stdout),
        ]
```
will bring the logging statements back when running `postprocess`.

kylebgorman: Right, that's intentional, the 2>/dev/null does that. I'll get rid of it though.

kylebgorman:
> This looks good, just a few comments.
> 
> Regarding logging, I like the idea of a 'structured report' mentioned in #156, but I think this ought to be fine. Perhaps if we direct the logging statements from `whitelist.py` to their own file (as is done with the logging statements generated in `src/scrape.py`) they will be easier to examine/work with?
> 
> Currently this is going to overwrite our old TSV's, which for the time being seems alright and `INPUT_TSV="${WHITELIST//whitelist/tsv}"` is an elegant solution. Eventually maybe we want to keep both the filtered and unfiltered data and give the filtered data a different name as mentioned in #155.

I think we just want to rewrite them; otherwise you have to reconfigure `./generate_summary.py` to handle the new paths. (It'll confuse "filtered" ones with dialect versions otherwise.)

data/wikipron/src/whitelist.py
kylebgorman: Not hiding stderr anymore so this should not be necessary.

PR #160, opened 2020-05-04
[quc] Allow space in K'iche' words.

### LINKED ISSUE #159: [quc] K'iche' support
K'iche' has [117 words with IPA pronunciations](https://en.wiktionary.org/wiki/Category:K%27iche%27_terms_with_IPA_pronunciation) but

    wikipron quc | wc -l

only gives us 65 entries. I believe this is because many of the entries have whitespace in the graphemic side and that we just need to enable 

        "no_skip_spaces_word": true,
        "no_skip_spaces_pron": true,

for this language.

kylebgorman:
This still falls just short of 100 (96) though there are 117 pronunciations according to Wiktionary.

Closes #159 (though not satisfactorily).

### DIFF
--- a/data/wikipron/src/languages.json
+++ b/data/wikipron/src/languages.json
@@ -440,7 +440,9 @@
         "iso639_name": "K'iche'",
         "wiktionary_name": "K'iche'",
         "wiktionary_code": "quc",
-        "casefold": true
+        "casefold": true,
+        "no_skip_spaces_word": true,
+        "no_skip_spaces_pron": true
     },
     "kbd": {
         "iso639_name": "Kabardian",


lfashby:
As far as I can tell this is working as expected. We [skip dashes](https://github.com/kylebgorman/wikipron/blob/master/wikipron/scrape.py#L22) in words and the 21 entries that we don't grab all contain dashes.

kylebgorman:
Ah, that's it. Okay if I leave this pending for a bit, in case somebody
comes through and adds enough pronunciations to get us over the 100-word
threshold?

kylebgorman:
Removing this; can revive if desired.

PR #163, opened 2020-05-05
[eng_uk] update

kylebgorman:
Run new scrape on eng_uk and tightens whitelist to reflect manual changes made to Wiktionary.


data/wikipron/whitelist/eng_uk_phonemic.whitelist
```diff
@@ -14,6 +14,7 @@ uː  #  Check.
 æ
 e
 o
+oː  # Check.
```
jacksonllee, line 17: What doe the comment "Check." mean for a given phone/phoneme?

kylebgorman:
In this particular case I have no idea why this exists, English doesn't
have length contrasts in any meaningful sense, yet it is in the data. Going
to cross-check against the transcription guidelines on Wiktionary to see if
there's any reason for this to exist, and if not, manually or automatically
correct.

PR #174, opened 2020-05-19
[uig] Split Uighur into Perso-Arabic and Cyrillic. 

agutkin:


data/wikipron/src/languages.json
```diff
@@ -1068,7 +1068,11 @@
         "iso639_name": "Uighur",
         "wiktionary_name": "Uyghur",
         "wiktionary_code": "ug",
-        "casefold": true
+        "casefold": true,
+        "script": {
+            "cyrl": "Cyrillic",
```
kylebgorman, line 1073: You can, if you want, leave off line 1073. As far as I know, the Cyrillic (and the Latin) alphabet were never widely used.

agutkin: Let's keep for now - who knows how many edits from Kazakhstan may enter the Wiktionary in the future.

kylebgorman: Roger that.

Issue #172, opened 2020-05-18
Apostrophes (0x27) not allowed in Latin-script spellings


agutkin:
Currently the regular expression in `generalized_split.py` behaves as follows for Maltese (mlt) (note, the following code is not exactly the same, but demonstrates the issue):

```
import regex
rs = r"^\p{Script=Latin}+$"
>>> bool(regex.match(rs, "jweġġa"))
True
>>> bool(regex.match(rs, "jweġġa'"))
False
```

This means that setting the script to Latin will filter out all the ASCII apostrophes.

According to [Quora](https://www.quora.com/In-Maltese-linguistics-what-do-the-appeared-apostrophes-after-many-of-the-words-mean-Is-this-some-sort-of-diacritic-mark):

> "If you mean the apostrophe in words such as ta’ or seba’, the apostrophe marks a place
> where there was historically a għ sound, which dropped off completely in that position in
> pronunciation but is still ‘remembered’ in writing through the apostrophe.
> 
> It is thus a little bit similar to the circumflex accent in French words that ‘remembers’ the
> place where Latin used to have an -s- sound which was dropped in pronunciation.
> 
> Note that often, Maltese words ending in an apostrophe may have related words where the għ
> is still present (because it is no longer at the end of the word), as with tagħna or sebgħa."

kylebgorman:
I guess we have to do something like:

    r"^[\p{Script=Latin}\p{Punct}]" 

or whatever.

agutkin:
And I suppose this logic has to be language-specific, possibly controlled by a flag specified in json config.

kylebgorman:
I see why you say that, but no, I'm not sure it does. We could just let any
(ASCII) punctuation chars through, I suppose. Or we could just let ' (which
is common in Latin scripts) through globally.

kylebgorman:
What I'm proposing is that you modify [generalized_split.py](https://github.com/kylebgorman/wikipron/blob/master/data/wikipron/src/generalized_split.py#L14) so that line 14 reads:

    regex_string = rf"^(\p{{{prop}={script}}}|\p{Punct})+$"

That should be all that's required. Would be nice to have a bigger Maltese dictionary!

Issue #166, opened 2020-05-16
[mlt] Blacklist entries in Arabic script?


agutkin:
Maltese lexicon has one entry in Arabic script. I suspect it's ended there from an etymology section somewhere. Is there a way to blacklist scripts during scraping?

kylebgorman:
If you add an inner "scripts" dictionary with `"ara"` and `latn` entries, it'll have that effect.

agutkin:
The issue is that Maltese should not have entries in Arabic AFAIK.

kylebgorman:
It'll have that effect. Or if you want to be totally explicit you can say:

   "scripts": { "latn": "Latn" }

We also do something similar, IIRC, to filter out a handful of non-Han entries under Mandarin.

agutkin:
This also drops all the spellings containing the apostrophe (').

kylebgorman:
Oh interesting. This is the ASCII one (0x27)?

If so let's make another bug to track that down. I would have thought
`generalized_split.py` would pass through punctuation (but not space or
dash), but if not it should be easy to add in.

agutkin:
Opened #172.

PR #175, opened 2020-05-21
[mlt] Latin-only spellings and allow for apostrophes.

agutkin:
Note: I followed a more conservative route by allowing a single symbol rather than punctuation universe.

#166 #172


kylebgorman:
Closes #172.

PR #180, opened 2020-05-29
[mon] Split Mongolian into Cyrillic script and Old Mongol script.

agutkin:
Note: mon_mong_{phonemic,phonetic}.tsv has 7 and 2 prons, respectively.


data/wikipron/src/languages.json
```diff
@@ -693,7 +693,11 @@
         "iso639_name": "Mongolian",
         "wiktionary_name": "Mongolian",
         "wiktionary_code": "mn",
-        "casefold": true
+        "casefold": true,
+        "script": {
+            "cyrl": "Cyrillic",
+            "mong": "Mong"
```
kylebgorman, line 699: Since it looks like there's not nearly enough of this you could remove the Mong line?

agutkin: Sure. Removed the old script from the registry.

Issue #149, opened 2020-04-18
Dialect support in extraction functions


lfashby:
As mentioned in #146, certain of our extraction functions do not support dialects. 

Currently, dialects are handled by the `_get_pron_xpath_selector` method in `config.py`, the results of which are stored in `pron_xpath_selector`. [`_yield_phn` in our default extraction function ](https://github.com/kylebgorman/wikipron/blob/master/wikipron/extract/default.py#L22) makes use of `config.pron_xpath_selector`, as does [Japanese](https://github.com/kylebgorman/wikipron/blob/master/wikipron/extract/jpn.py#L41) (because its extraction function is used for targeting orthographic entries). 

Our Chinese, Khmer, Latin, and Thai extraction functions do not use the default pron selector and skip over `_yield_phn` by calling [`yield_pron`](https://github.com/kylebgorman/wikipron/blob/master/wikipron/extract/core.py#L23) directly - meaning dialect information is ignored.

jacksonllee:
@kylebgorman @lfashby Should we close this issue? IIRC, we opened this ticket due to not having been able to scrape for Latin by its dialects -- now solved by #233 (Kyle and I stepped through the Latin HTML on Wiktionary together and realized the solution would actually be a lot simpler than we might have expected). Resolving the Latin has led me to think if a language has its custom extraction function and needs dialect support, then it will probably need its own custom solution for dialects as well. There's really no way we can have a generalized solution for dialect specification for languages with custom extraction functions. We can always file more specific tickets for specific languages.

kylebgorman:
+1

lfashby:
I suppose it’s fine to close this. Do we need to make it known somewhere that our `--dialect` flag won't do anything for (certain) languages with extraction functions?

jacksonllee:
> Do we need to make it known somewhere that our --dialect flag won't do anything for (certain) languages with extraction functions?

I don't have a very strong preference, but I feel like this ticket has been serving this purpose, with the problem that this ticket isn't really actionable unless there's another more specific, probably language-particular ticket (like the Latin one). This being said, I could see perhaps a ticket like this is the only sensible thing to do for visibility in our current set-up, given that it's mostly only for developers and would be meaningless in any user-facing documentation for others (e.g. what are the "extraction functions"?).

I'm getting through a couple (known) issues and will update my thoughts here.

kylebgorman:
[...]

PR #181, opened 2020-06-10
Vietnamese extraction function

### LINKED ISSUE #133: [vie] Enhancements

lfashby:

This PR includes an extraction function for Vietnamese, data from scraping Vietnamese with this new extraction function, and the removal of a duplicate of the Manchu entry in `languages.json` (I'll have to look into how that got there). 

This Vietnamese extraction function is tentative and works around the problem raised in #149 by redefining and utilizing  `_DIALECT_XPATH_SELECTOR_TEMPLATE` within the extraction function itself. Whatever solution we come up with to handle #149 will likely require reworking this extraction function, but I've submitted it as is because it makes clear some of the problems we will face in solving #149 - such as that Vietnamese dialects require a different dialect selector than is needed for other languages and that extraction functions which do not call `_yield_phn` need some way of being exposed to the dialect specifications.

Resolves #133


kylebgorman:
Gentle ping: should I merge this?

lfashby:
I'll merge it in. I reached out to Jackson separately regarding this.

PR #184, opened 2020-06-26
Merged Whitelist functionality with src/scrape.py. Now checks for pre…

bonham79:
…sence of whitelist and writes separate tsv as {original file name}_filtered.tsv. Update generate_summary to reflect if file is filtered through a whitelist. CHANGELOG and README update accordingly.

- [x] Updated `Unreleased` in `CHANGELOG.md` to reflect the changes in code or data.  #155

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,9 +29,13 @@ Unreleased
 -   Alowing ASCII apostrophes (0x27) in spellings. (\#172).
 -   Only allowing Latin spellings in Maltese lexicon. (\#166).
 -   Split `mon` into Cyrillic and Mongol Bichig (\#179).
--   Added Vietnamese extraction function. (\#181)
--   Modified pron selector in Latin extraction function. (\#183)
-
+-   Added Vietnamese extraction function. (\#181).
+-   Modified pron selector in Latin extraction function. (\#183).
+-   Merged whitelist.py into 'big scrape' script. src scrape.py now checks for
+    existence of whitelist file during scrape to create second 'filtered' tsv.
+    New tsv stored in tsv folder under {"default tsv filename"}_filtered.tsv.
+    (\#154).
+-   Updated generate_summary to reflect presence of 'filtered' tsv. (\#154)
 ### Deprecated
 ### Removed
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,9 +32,9 @@ Unreleased
 -   Added Vietnamese extraction function. (\#181).
 -   Modified pron selector in Latin extraction function. (\#183).
 -   Merged whitelist.py into 'big scrape' script. src scrape.py now checks for
-    existence of whitelist file during scrape to create second 'filtered' tsv.
-    New tsv stored in tsv folder under {"default tsv filename"}_filtered.tsv.
-    (\#154).
+    existence of whitelist file during scrape to create second filtered TSV.
+    New TSV is placed under tsv/\*\_filtered.tsv. (\#154).
+
 -   Updated generate_summary to reflect presence of 'filtered' tsv. (\#154)
 ### Deprecated
 ### Removed
--- a/data/wikipron/src/README.md
+++ b/data/wikipron/src/README.md
@@ -3,14 +3,14 @@
 
 [`scrape.py`](scrape.py) calls WikiPron's scraping functions on all Wiktionary
 languages with over 100 entries. If a whitelist entry is present for language,
-will generate an extra 'filtered' file only containing whitelist phones/phonemes.
-[`generate_summary.py`](generate_summary.py) generates a [README](../README.md)
-and a [TSV](../languages_summary.tsv) with selected information regarding the
-contents of the TSVs  [`scrape.py`](scrape.py) generated and the configuration
-settings that were passed to scrape. [`postprocess`](postprocess) sorts and
-removes entries in each TSV if they have the same graphemic form and 
-phonetic/phonemic form as a previous entry. In addition it splits TSVs
-containing multiple scripts (Arabic, Cyrillic, etc.)
+will generate an additional file filtered to only containing whitelist
+phones/phonemes.[`generate_summary.py`](generate_summary.py) generates a
+[README](../README.md) and a [TSV](../languages_summary.tsv) with selected
+information regarding the contents of the TSVs [`scrape.py`](scrape.py)
+generated and the configuration settings that were passed to scrape.
+[`postprocess`](postprocess) sorts and removes entries in each TSV if they have
+the same graphemic form and phonetic/phonemic form as a previous entry. In
+addition it splits TSVs containing multiple scripts (Arabic, Cyrillic, etc.)
 into constituent TSVs containing a single script.
 [`languages.json`](languages.json) provides [`scrape.py`](scrape.py) with a
 dictionary containing the information it needs to call scrape on all Wiktionary
@@ -28,10 +28,10 @@ Steps used to update the dataset
 1.  Run [`codes.py`](codes.py) to update [`languages.json`](languages.json).
     -   If there are new Wiktionary languages with over 100 entries, they will
         be added to [`languages.json`](languages.json).
-    -   As mentioned in the comment at the top of [`codes.py`](codes.py), any new
-        languages added to [`languages.json`](languages.json) will not have their
-        case-folding settings specified. Their `"casefold"` value will therefore
-        be set to `None`.
+    -   As mentioned in the comment at the top of [`codes.py`](codes.py), any
+        new languages added to [`languages.json`](languages.json) will not have
+        their case-folding settings specified. Their `"casefold"` value will
+        therefore be set to `None`.
     -   Whether or not to apply case-folding for these new languages needs to be
         manually set by changing the `"casefold"` value within
         [`languages.json`](languages.json).
@@ -41,11 +41,12 @@ Steps used to update the dataset
     -   By default `cut_off_date` in `main()` is set using
         `datetime.date.today().isoformat()` but can be set manually using an ISO
         formatted string (ex. "2019-10-31").
-3.  Run [`postprocess`](postprocess).
+3.  Run [`postprocess`](postprocess).filtered
 4.  Run [`generate_summary.py`](generate_summary.py).
 
 Running a subset of languages using the big scrape
 --------------------------------------------------
+
 By default, if [`scrape.py`](scrape.py) cannot successfully complete a scrape of
 an entire language in 10 retries, it will log the language, remove the
 incomplete data scraped from that language, and move on to the next language in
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -29,9 +29,13 @@ Unreleased
...
 -   Only allowing Latin spellings in Maltese lexicon. (\#166).
 -   Split `mon` into Cyrillic and Mongol Bichig (\#179).
--   Added Vietnamese extraction function. (\#181)
--   Modified pron selector in Latin extraction function. (\#183)
-
+-   Added Vietnamese extraction function. (\#181).
+-   Modified pron selector in Latin extraction function. (\#183).
+-   Merged whitelist.py into 'big scrape' script. src scrape.py now checks for
+    existence of whitelist file during scrape to create second 'filtered' tsv.
+    New tsv stored in tsv folder under {"default tsv filename"}_filtered.tsv.
```
kylebgorman, line 36: "TSV" in prose (it's an acronym), thus:

   The new TSV file is placed under `tsv/*_filtered.tsv`.

CHANGELOG.md
```diff
@@ -29,9 +29,13 @@ Unreleased
...
+    (\#154).
+-   Updated generate_summary to reflect presence of 'filtered' tsv. (\#154)
```
kylebgorman, line 38: ...filtered TSV.

data/wikipron/src/README.md
```diff
@@ -2,13 +2,15 @@
...
 [`scrape.py`](scrape.py) calls WikiPron's scraping functions on all Wiktionary
-languages with over 100 entries. [`generate_summary.py`](generate_summary.py)
-generates a [README](../README.md) and a [TSV](../languages_summary.tsv) with
-selected information regarding the contents of the TSVs [`scrape.py`](scrape.py)
-generated and the configuration settings that were passed to scrape.
-[`postprocess`](postprocess) sorts and removes entries in each TSV if they have
-the same graphemic form and phonetic/phonemic form as a previous entry. In
-addition it splits TSVs containing multiple scripts (Arabic, Cyrillic, etc.)
+languages with over 100 entries. If a whitelist entry is present for language,
+will generate an extra 'filtered' file only containing whitelist phones/phonemes.
```
kylebgorman, line 6: Not sure why 'filtered' is quoted here...the meaning is clear, and I don't think you're doing any sort of ironic distance ;)

data/wikipron/src/README.md
```diff
@@ -2,13 +2,15 @@
 ====================
```
kylebgorman, line 2: Can you reflow this file? I use pandoc for this:

    pandoc -f markdown -t markdown source tm; mv tmp source

or something like that.

data/wikipron/src/scrape.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
kylebgorman, line 1: Is this change intentional? (For me `python` is an alias for `python3`.)

data/wikipron/src/scrape.py
```diff
@@ -18,23 +18,57 @@
...
             try:
-                for (word, pron) in wikipron.scrape(config):
-                    print(f"{word}\t{pron}", file=source)
+                scrape_results = wikipron.scrape(config)
+                # Given whitelist, opens up a second tsv for scraping.
+                if whitelist_set:
+                    with open(tsv_path_filtered, "w") as source_filtered:
+                        for (word, pron) in scrape_results:
+                            if _filter(word, pron, whitelist_set):
+                                print(f"{word}\t{pron}", file=source_filtered)
```
kylebgorman, line 57: This is pretty minor but rather than building this string to print twice, I would maybe do something like:

    line = f"{word}\t{pron}
    if _filter(...):
        print(line, file=...)
    print(line)

data/wikipron/src/scrape.py
```diff
@@ -18,23 +18,57 @@
...
+        return True
+
+
 def _call_scrape(
-    lang_settings: Dict[str, str], config: wikipron.Config, tsv_path: str
+    lang_settings: Dict[str, str],
+    config: wikipron.Config,
+    tsv_path: str,
+    whitelist_set: FrozenSet[str] = None,
+    tsv_path_filtered: str = "",
```
kylebgorman, line 46: Maybe just call this `tsv_filtered_path` instead?

data/wikipron/src/scrape.py
```diff
@@ -45,19 +79,65 @@ def _call_scrape(
...
 
 def _build_scraping_config(
     config_settings: Dict[str, Any], wiki_name: str, dialect_suffix: str = ""
 ) -> None:
     path_affix = f'../tsv/{config_settings["key"]}_{dialect_suffix}'
+    whitelist_path_affix = (
+        f"../whitelist/{config_settings['key']}_{dialect_suffix}"
+    )
+
+    # Configures phonemic TSV
```
kylebgorman, line 96: Period at end of comment (and below) please. (I don't have a style guide to point to this but it's how we've done things.)

lfashby:
This is a neat way of utilizing the whitelists. 

Don't these changes require a modification of our `postprocess` script? The whitelist section of `postprocess` is now redundant as far as I can tell.
#155 indicates that we want to keep both the unfiltered and filtered TSVs. Is that still our desire? I ran these changes locally and found that `generate_summary.py` can't handle the new filtered file names. There are several options of ways to fix that problem, but even when fixed it leaves the question of how the filtered files should be labelled in the README table of the TSVs so that they can be differentiated from the regular files. Perhaps another column indicating whether or not filtering has been applied?

One of the motivations (at least from a few months) behind incorporating a whitelisting value or something similar into entries in `languages.json` was so that `generate_summary.py` could handle filtered files in a similar fashion to how dialects are handled. Perhaps we don't need to go with that approach, but we do, I think, need a way of differentiating the `eng_uk_phonemic_filtered.tsv` row from the `eng_uk_phonemic.tsv`row in the TSV table README.

bonham79:
What outputs were you getting for filtered files? generate_summary should ideally be outputting along the lines of "phonemic_filtered" for its transcription value (not the cleanest but seemed to have least impact on the script). Is this not the case?

lfashby:
Sorry, probably should have explained the error I got. Running `generate_summary.py` with a file like `eng_us_phonemic_filtered.tsv` in the `tsv` directory will lead to a key error on [line 31](https://github.com/kylebgorman/wikipron/blob/75339520aaf78e4b151d7794dde277ba37a266ed/data/wikipron/src/generate_summary.py#L31) because the key isolated from line 23 will be `us_phonemic`.
I overlooked the modifications to `transcription_level`. I think that should adequately differentiate the files in the README table. Maybe not the cleanest solution, but certainly the easiest, which is fine with me.

bonham79:
Ah, I see. I was error testing on the Greek so didn't see dialect issues. Is there any longterm plan/design philosophy for the indexing to not simply look for "phone" instead of a position underscore? It seems the most reliable for file naming conventions but not sure if that was intentionally avoided.

kylebgorman:
Let's make sure both of them (filtered and unfiltered) end up in the table, by whatever means necessary. I guess the file name parsing needs to be slightly more robust to handle dialect, filtered/not, and script.

lfashby:
I think we went with indexing the underscores because originally all the files were just `isokey_phonetic.tsv` or `isokey_phonemic.tsv`. We probably should have changed things when more underscores were added.

One other note, I haven't tested this but I'd imagine the new filtered file names will break `generalized_split.py` as well (more [underscore indexing](https://github.com/kylebgorman/wikipron/blob/04130daf743ccd1f39d968c4b3662b0bab0a20a2/data/wikipron/src/generalized_split.py#L34)).

bonham79:
Alright I'm just going to push the current changes for now and then start on looking at the issue cascade then.

bonham79:
So it looks like generalized_split was robust enough already. Just needed a tweak as far as I can tell. And whitelist can be removed for redundancies. May need a more in-depth fix for generate_summary and generalized_split where postioning of file name is less important for quality of life? (Thinking just have them load language codes (dialect, filtered, ISO) and search for their regex values without concern for order. Probably unnecessary but could avoid future issues.)

data/wikipron/src/postprocess
```diff
@@ -10,16 +10,6 @@ split() {
...
-        OUTPUT_TSV=$(mktemp)
-        # TODO: Redirect these reports somewhere sensible.
-        ./whitelist.py "${INPUT_TSV}" "${WHITELIST}" "${OUTPUT_TSV}"
-        mv "${OUTPUT_TSV}" "${INPUT_TSV}"
-    done
-}
-
 main() {
     time split
     time whitelist
```
lfashby, line 15: I think we can get rid of this.

data/wikipron/src/scrape.py
```diff
@@ -18,46 +18,127 @@
...
         with open(tsv_path, "w") as source:
             try:
-                for (word, pron) in wikipron.scrape(config):
-                    print(f"{word}\t{pron}", file=source)
+                scrape_results = wikipron.scrape(config)
+                # Given whitelist, opens up a second tsv for scraping.
+                if whitelist_set:
+                    with open(tsv_filtered_path, "w") as source_filtered:
+                        for (word, pron) in scrape_results:
+                            line = f"{word}\t{pron}
```
lfashby, line 56: Need a closing `"` here?

lfashby:
Other than those two things above, looks great to me.

data/wikipron/src/scrape.py
bonham79: Sorry about that. My text editor is being finicky and didn't register that edit.

data/wikipron/src/README.md
```diff
@@ -39,11 +41,12 @@ Steps used to update the dataset
     -   By default `cut_off_date` in `main()` is set using
         `datetime.date.today().isoformat()` but can be set manually using an ISO
         formatted string (ex. "2019-10-31").
-3.  Run [`postprocess`](postprocess).
+3.  Run [`postprocess`](postprocess).filtered
```
kylebgorman, line 44: typo here?

data/wikipron/src/scrape.py
```diff
@@ -18,46 +18,127 @@
...
-    # Log and remove TSVs for languages that failed
+    # Log and remove TSVs for languages that failed.
     # to be scraped within 10 retries.
     logging.info(
         'Failed to scrape "%s" within 10 retries. %s',
         lang_settings["key"],
         lang_settings,
     )
+    # Checks if second TSV was opened.
+    if os.path.exists(tsv_filtered_path):
```
kylebgorman, line 84: Suggest using try/except here instead of testing for existence. In theory it can help you avoid a weird race condition.

data/wikipron/src/README.md
bonham79: Definitely. I'm never using atom again.

data/wikipron/src/scrape.py
bonham79: Could you go into more detail? I didn't know that could happen here.

kylebgorman: There's no particular reason to think it would occur here but the file could *in theory* be materialized between line 84 and line 85, in which case it would not be removed, or it could be deleted between lines 84 and 85, in which case this would throw an uncaught exception. So the general recommendation is try and fail.

bonham79:
> Do you also want to remove `whitelist.py` in `src/`?

I'm thinking it can still be left in for testing purposes? Just to check if the whitelists being passed are giving effective results without needing to call a scrape.

kylebgorman:
Okay, fine with me to leave `whitelist.py` in there.

PR #188, opened 2020-07-13
Add --no-tone flag

lfashby:

This PR offers a potential solution to issue #91. It seems to work as desired and removes all the tones mentioned by Jackson in #91.The range of tones removed is demonstrated in the changes to `test_config.py`. 
The [regex used to target the various tones](https://github.com/lfashby/wikipron/blob/9001fe5317f2c4b0d819dab53184894e2c225f62/wikipron/config.py#L39) we want to remove is not particularly 'legible'. Perhaps I should add comments detailing the character each codepoint refers to? Or I could replace the codepoints with characters? If anyone has any other suggestions let me know.

### DIFF
--- a/wikipron/config.py
+++ b/wikipron/config.py
@@ -2,6 +2,7 @@ import datetime
 import functools
 import logging
 import re
+import unicodedata
 
 from typing import Callable, Optional
 
@@ -35,6 +36,11 @@ _DIALECT_XPATH_SELECTOR_TEMPLATE = (
 _PHONEMES_REGEX = r"/(.+?)/"
 _PHONES_REGEX = r"\[(.+?)\]"
 
+_TONES_REGEX = (
+    r"[\u02E5-\u02E9\u00b2-\u00b3\u00b9\u2074-\u2079"
+    r"\u030B\u030C\u030F\u0300-\u0302\u0304\u1DC4-\u1DC9\u207B-\u207E]"
+)
+
 
 class Config:
     """Configuration for a scraping run.
@@ -57,13 +63,14 @@ class Config:
         phonetic: bool = False,
         dialect: Optional[str] = None,
         no_segment: bool = False,
+        no_tone: bool = False,
         no_skip_spaces_word: bool = False,
         no_skip_spaces_pron: bool = False,
     ):
         self.language: str = self._get_language(key)
         self.casefold: Callable[[Word], Word] = self._get_casefold(casefold)
         self.process_pron: Callable[[Pron], Pron] = self._get_process_pron(
-            no_stress, no_syllable_boundaries, no_segment
+            no_stress, no_syllable_boundaries, no_segment, no_tone
         )
         self.cut_off_date: str = self._get_cut_off_date(cut_off_date)
         self.ipa_regex: str = _PHONES_REGEX if phonetic else _PHONEMES_REGEX
@@ -120,13 +127,20 @@ class Config:
         return str.casefold if casefold else lambda word: word  # noqa: E731
 
     def _get_process_pron(
-        self, no_stress: bool, no_syllable_boundaries: bool, no_segment: bool
+        self,
+        no_stress: bool,
+        no_syllable_boundaries: bool,
+        no_segment: bool,
+        no_tone: bool,
     ) -> Callable[[Pron], Pron]:
         processors = []
         if no_stress:
             processors.append(functools.partial(re.sub, r"[ˈˌ]", ""))
         if no_syllable_boundaries:
             processors.append(functools.partial(re.sub, r"\.", ""))
+        if no_tone:
+            processors.append(functools.partial(unicodedata.normalize, "NFD"))
+            processors.append(functools.partial(re.sub, _TONES_REGEX, ""))
         if not no_segment:
             processors.append(
                 functools.partial(segments.Tokenizer(), ipa=True)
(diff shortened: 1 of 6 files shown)


wikipron/config.py
```diff
@@ -120,13 +127,20 @@ def _get_casefold(self, casefold: bool) -> Callable[[Word], Word]:
...
+        no_segment: bool,
+        no_tone: bool,
     ) -> Callable[[Pron], Pron]:
         processors = []
         if no_stress:
             processors.append(functools.partial(re.sub, r"[ˈˌ]", ""))
         if no_syllable_boundaries:
             processors.append(functools.partial(re.sub, r"\.", ""))
+        if no_tone:
+            processors.append(functools.partial(unicodedata.normalize, "NFD"))
```
kylebgorman, line 142: Why is the NFD conversion necessary here? 

So if this flag enabled our Korean data comes out one _jamo_ per codepoint, but if it's not it comes out one _hangul_ per codepoint. That seems suboptimal to me.

lfashby: The NFD conversion is to handle the accents that Jackson referred to such as those used [here](https://en.wiktionary.org/wiki/beseni#Chichewa). `_TONES_REGEX` targets several of these combining diacritics, though thus far the ‘acute’ accent is the only one I’ve found used in transcriptions. 
I was wondering whether it was appropriate to do this conversion and am not sure if it will have any undesirable side effects, but since `segments.Tokenizer()` does it I assumed it would be okay to do it here as well.

kylebgorman: I didn't know that about `segments` applying NFD.

Let me suggest that we apply NFD to all pronunciations, immediately, rather than piecemeal (if segmentation and/or no-tones are enabled) before the configurable processing begins, presumably not in the config but earlier.

NFD conversion is _idempotent_ (is that the right word? a second application does nothing, in other words), so a second application by `segments` is harmless, and fast.

I wonder if we also want to apply NFC at the very end, also, because I consider that to be the standard representation for Unicode data at rest, and the graphemes are all NFC because Wiktionary (and the vast majority of web text in general) is all NFC. 

Sorry to add mission creep to this but I think those two steps (NFD on input, NFC on output) would be wise, WYDT?

wikipron/config.py
```diff
@@ -35,6 +36,11 @@
 _PHONEMES_REGEX = r"/(.+?)/"
 _PHONES_REGEX = r"\[(.+?)\]"
 
+_TONES_REGEX = (
+    r"[\u02E5-\u02E9\u00b2-\u00b3\u00b9\u2074-\u2079"
```
kylebgorman, line 40: Sorry if I made this comment earlier but is it possible to replace these Unicode numerical codepoints with their human-readable versions? I believe it'll work just fine. (For instance, to obtain the printable version of \u02E5 I do:

    chr(0x02E5)

wikipron/config.py
```diff
@@ -35,6 +36,8 @@
 _PHONEMES_REGEX = r"/(.+?)/"
 _PHONES_REGEX = r"\[(.+?)\]"
 
+_TONES_REGEX = r"[˥˦˧˨˩⁰¹²³⁴⁵⁶⁷⁸⁹⁻⁽⁾◌̋ ◌̌ ◌̏ ◌̀ ◌́ ◌̂ ◌̄ ◌᷄◌᷅◌᷆◌᷇◌᷈◌᷉]"
```
kylebgorman, line 39: This looks relatively complete except that I don't see the markers for downstep/upstep or for "global rise" and "global fall". Don't know if those are actually present in any of the transcripts but might be wise to include.
Screenshot 2020-07-22 at 10 43 50 AM

wikipron/config.py
```diff
@@ -120,13 +124,20 @@ def _get_casefold(self, casefold: bool) -> Callable[[Word], Word]:
...
+        no_segment: bool,
+        no_tone: bool,
     ) -> Callable[[Pron], Pron]:
         processors = []
         if no_stress:
             processors.append(functools.partial(re.sub, r"[ˈˌ]", ""))
         if no_syllable_boundaries:
             processors.append(functools.partial(re.sub, r"\.", ""))
+        if no_tone:
+            processors.append(functools.partial(unicodedata.normalize, "NFD"))
```
kylebgorman, line 139: This line 139 is no longer necessary given what you've done. It ought to be no-op.

wikipron/extract/core.py
```diff
@@ -35,7 +36,7 @@ def yield_pron(
         if _skip_pron(pron, config):
             continue
         try:
-            pron = config.process_pron(pron)
+            pron = config.process_pron(unicodedata.normalize("NFD", pron))
```
kylebgorman, line 39: Maybe do this in two steps:

   pron = unicodedata.normalize("NFD", pron)

Don't know why but I think that might be easier to read.

wikipron/scrape.py
```diff
@@ -41,8 +42,9 @@ def _scrape_once(data, config: Config) -> Iterator[WordPronPair]:
         ):
             continue
         request = session.get(_PAGE_TEMPLATE.format(word=word), timeout=10)
+        word = unicodedata.normalize("NFD", word)
         for word, pron in config.extract_word_pron(word, request, config):
-            yield word, pron
+            yield word, unicodedata.normalize('NFC', pron)
```
kylebgorman, line 47: This looks good to me but use double-quotes here.

kylebgorman: Also, could you add a line of documentation here:

# Pronunciation processing is done in NFD-space; we convert back to NFC afterwards.

lfashby:
Not sure if that commit description makes sense - the latest changes (as indicated by the updates to `test_config.py`) will have `--no-tone` avoid removing superscript parentheses around superscript characters that do not indicate tones. It will still remove superscript parentheses around characters that do indicate tones.

PR #205, opened 2020-09-07
Updates to codes.py

lfashby:

`codes.py` has been updated so that it is hopefully a bit easier to follow and doesn’t needlessly iterate through a large TSV of iso639 codes. 

`codes.py` and `test_languagecodes.py` had certain steps in common (and still have a bit in common now that `codes.py` tells you if WikiPron can’t handle any language codes used in the big scrape). To  reduce a bit of the previous redundancy between `codes.py` and `test_languagecodes.py` I appropriated the Wiktionary querying functions from `test_languagecodes.py`, placed them in `codes.py`, and imported them from there back into `test_languagecodes.py`. If that seems like an unnecessary work around to try and reduce a bit of repeated code or if you object to importing functions from the `data` directory into the `tests` directory let me know.


data/src/codes.py
```diff
@@ -8,182 +8,193 @@
...
+    """Checks if WikiPron can handle the assigned ISO language code."""
+    try:
+        language_inferred = wikipron.Config(key=language_code).language
+    except iso639.NonExistentLanguageError:
+        logging.warning('WikiPron cannot handle "%s"', language)
+    else:
+        if language_inferred != language:
+            logging.warning(
+                'WikiPron resolves the key "%s" to "%s"'
+                'which is listed as "%s" on Wiktionary.',
```
kylebgorman, line 129: Small nit but a quick search said most logging statements here don't end in '.', maybe use that.

Also instead of typing "%s" in a logging statement, use %r, which is the same thing. Then you can use double-quotes for the actual logging string, like the rest of the file.

kylebgorman (approved):
No, I think importing it like that makes a lot of sense.

PR #207, opened 2020-09-20
Rename "whitelist"

### LINKED ISSUE #206: Rename "whitelist"
Reason should be obvious. I would suggest:

* delete `whitelist.py`; we don't need it anymore
* rename the file extension `.phones`
* rename the subdirectory `phones`

jacksonllee:
Resolves #206


jacksonllee:
@kylebgorman @lfashby Is there a way we could check that these changes haven't broken anything without triggering the "big scrape" process?

kylebgorman:
You could run the actual scrape (`src/./scrape.py`) through Armenian (which will at most take a few hours) and confirm there are no obvious regressions, then revert.

kylebgorman (approved):
LGTM

lfashby:
My only suggestion would be to run a partial big scrape and then just manually compare the data. It would be nice if we had a way of testing some of these things on the `data` side of WikiPron.

PR #243, opened 2020-10-20
[kmr] New scrape: Northern Kurdish

agutkin:
It's not entirely clear to me whether this should go under general Kurdish (`kur`) as a separate dialect or be kept as is. I guess it makes sense to keep it separate for now.


kylebgorman:
I'm told Kurmanji (as this one is called) and Sorani are quite different. If they have separate ISO codes and they're encoded separately on Wiktionary, that's a strong clue IMO.

kylebgorman (approved):
LGTM

PR #245, opened 2020-10-20
Add logging for dialect support for languages requiring custom extraction logic

### LINKED ISSUE #149: Dialect support in extraction functions

jacksonllee:
This PR adds logging to warn about dialect support for languages that require custom extraction logic. As a test using Latin:

```
$ wikipron --dialect=classical lat
INFO: Language: 'Latin'
INFO: No cut-off date specified
INFO: Dialect(s): 'classical'
INFO: 'Latin' requires custom logic to handle its data from Wiktionary. The dialect parameter, specified for 'classical', may or may not work as desired. If you notice any issues, please report them at https://github.com/kylebgorman/wikipron/issues.
```

Closes #149.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,8 @@ Unreleased
 -   Added HTTP User-Agent header to API calls to Wiktionary. (\#234)
 -   Added support for python 3.9 (\#240)
 -   Added black style formatting to `.circleci/config.yml`. (\#242)
+-   Added logging for scraping a language with `--dialect` specified
+    that requires its custom extraction logic. (\#245)
 
 #### Changed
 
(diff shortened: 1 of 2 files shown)


kylebgorman (approved):
Message looks good! LGTM

Issue #235, opened 2020-10-15
Repo maintenance and upgrades, 2020 edition


jacksonllee:
Placeholder ticket for a couple repo maintenance tasks we should get to before we make a new release:

* Turn on Python 3.9 support (turn on CircleCI builds for this version, add trove classifier in setup.py)
* Consider adding `black --line-length=79 wikipron tests data` to CircleCI build config to enforce style consistency
* The changelog needs organizing. Consider separating the entries into two categories of something like "data" and "WikiPron package and scraping tools". Update the PR template markdown accordingly.
* Add missing entries to the changelog, if any.
* Check dependency versions (both dev and non-dev packages). Bump versions if appropriate.

kylebgorman:
* I would love also to enable `mypy` and `flake8` on CircleCI. One issue with `flake8` is that it disagrees with `black` on one annoying issue: `black` adds spaces between : slices, and `flake8` complains about them.
* I agree a data sectio nfor the changelog would be wise.
* What about just making a release? We haven't done one in a while.

jacksonllee:
I'm going to write individual, actionable GitHub tickets for the tasks listed in this current ticket. Apparently, it's Hacktoberfest month now and we can label the appropriate tickets with "good first issue" and "hacktoberfest" for anyone looking to contribute, and I'm going to use these labels. This current ticket is meant to be just a discussion thread to collect ideas, not an actionable ticket to be closed by code changes.

* flake8 vs black: I had also run into them not playing nice together recently for other reasons. Though I vaguely remember once all these packages were updated to their latest versions, they appeared to work well together? We will see.
* mypy: I remember we had this conversation before offline. If (the latest version of) mypy doesn't give us more trouble than otherwise, I'd be happy to see it applied in this repo.

PR #247, opened 2020-10-23
ISSUE-241: Use mypy for type checking

benfernandes:
#241 

Tested this locally using the contribution guidelines. MyPy, Flake8, Black and Pytest all passed.

PR #249, opened 2020-10-24
Improve CircleCI workflow

platipo:

Hi @jacksonllee, I saw #235 and decided to hop on the renovation bandwagon for he CircleCI job. I'm working on the [Orbtoberfest](https://hacktoberfest.circleci.com/) challenges and saw an opportunity to improve the actual configurations:
- [x] add the [matrix](https://circleci.com/blog/circleci-matrix-jobs/) building syntax for improved readability
- [ ]  add parallel testing for faster pipelines
- [x] add Python orb to remove maintenance burden
- [x] cache pipeline dependencies

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -81,6 +81,10 @@ workflows:
           name: black
           command-name: black
           command-run: black --line-length=79 --check project/setup.py project/wikipron project/tests project/data
+      - pre-build:
+          name: mypy
+          command-name: mypy
+          command-run: mypy --ignore-missing-imports project/wikipron project/tests project/data
       - build-python:
           requires:
             - flake8
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,11 +41,6 @@ jobs:
         #  password: $DOCKERHUB_PASSWORD
     steps:
       - checkout
-      - run:
-          name: Create empty pyenv file
-          command: |
-              mkdir -p /home/circleci/.pyenv/versions/
-              touch /home/circleci/.pyenv/version
       - python/install-packages:
           pkg-manager: pip
           pre-install-steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,9 +14,12 @@ jobs:
       - checkout
       - python/install-packages:
           pre-install-steps:
-            - pip install --progress-bar off --upgrade pip setuptools
-            - python setup.py sdist
-            - pip install dist/*.tar.gz
+              - run:
+                  name: Build source distribution and install package from it
+                  command: |
+                      pip install --progress-bar off --upgrade pip setuptools
+                      python setup.py sdist
+                      pip install dist/*.tar.gz
       - run:
           name: Show installed Python packages
           command: pip list
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,7 @@ jobs:
       command-run:
         type: string
     docker:
-      - image: cimg/python:3.7
+      - image: cimg/python:3.9
         #auth:
         #  username: $DOCKERHUB_USERNAME
         #  password: $DOCKERHUB_PASSWORD
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  python: circleci/python@0.3.2
+  python: circleci/python@1.2.0
 
 jobs:
   build-python:
@@ -12,24 +12,14 @@ jobs:
       - image: python:<< parameters.python-version >>
     steps:
       - checkout
-      - python/load-cache
-      - run:
-          name: Install Python build tools
-          command: |
-              pip install --progress-bar off --upgrade pip setuptools
-      - python/dist
-      - run:
-          name: Build source distribution and install package from it
-          # Ensure we can build a source distribution that can correctly install.
-          # "python setup.py sdist" creates dist/wikipron-x.y.z.tar.gz
-          command: |
-              pip install dist/*.tar.gz
-      - python/install-deps:
-          local: false
+      - python/install-packages:
+          pre-install-steps:
+            - pip install --progress-bar off --upgrade pip setuptools
+            - python setup.py sdist
+            - pip install dist/*.tar.gz
       - run:
           name: Show installed Python packages
           command: pip list
-      - python/save-cache
       - run:
           name: Type checking
           working_directory: ~/
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,7 +31,7 @@ jobs:
           command: pip list
       - run:
           name: Chown python packages
-          command: chown cicleci:circleci -R /usr/local/lib/python<< parameters.python-version >>/site-packages
+          command: chown circleci:circleci -R /usr/local/lib/python<< parameters.python-version >>/site-packages
       - persist_to_workspace:
               root: /usr/local/lib/python<< parameters.python-version >>
               paths:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -54,5 +54,5 @@ workflows:
     jobs:
       matrix:
         parameters:
-          node-version: ["3.6", "3.7", "3.8", "3.9"]
+          python-version: ["3.6", "3.7", "3.8", "3.9"]
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -73,14 +73,12 @@ workflows:
     jobs:
       - pre-test:
           name: pre-flake8
-          parameters:
-            command-name: flake8
-            command-run: flake8 project/setup.py project/wikipron project/tests
+          command-name: flake8
+          command-run: flake8 project/setup.py project/wikipron project/tests
       - pre-test:
           name: pre-black
-          parameters:
-            command-name: black
-            command-run: black --line-length=79 --check project/setup.py project/wikipron project/tests project/data
+          command-name: black
+          command-run: black --line-length=79 --check project/setup.py project/wikipron project/tests project/data
       - build-python:
           requires:
             - pre-flake8
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,9 +66,8 @@ jobs:
           command: |
               sudo chown circleci:circleci /home
               pytest -vv project/tests --junitxml /tmp/testxml/report.xml
-      - run: cat /tmp/testxml/report.xml
       - store_test_results:
-          path: /tmp/testxml/report.xml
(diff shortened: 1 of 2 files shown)


platipo:
One last point is left:

>  - [ ] add parallel testing for faster pipelines

In my opinion is not worth it to apply the [CircleCI parallelism](https://circleci.com/docs/2.0/parallelism-faster-jobs/) because there are only few tests and it would require splitting the build-python job into two halves: the former would build package ad its requirements while the latter would run the tests. This splitting is required in order to avoid unknown behavior in the cache save.

Let me know if you want to test it anyway.

platipo:
I reworked the entire workflow:

new workflow

If either black of flake8 fails the other jobs won't start. Now building the package and installing dependencies are cached so it needs less time to complete the run.

platipo:
Do I have to anything with ` Updated Unreleased in CHANGELOG.md to reflect the changes in code or data.`?

.circleci/config.yml
```diff
@@ -1,63 +1,91 @@
...
+  pre-build:
+    description: Install and run a Python standalone package
+    parameters:
+      command-name:
+        type: string
+      command-run:
+        type: string
     docker:
-      - image: python:3.6
+      - image: cimg/python:3.7
```
jacksonllee, line 16: If we're running flake8/black/mypy in just one Python version, I am wondering if we can use `cimg/python:3.9` here since Python 3.9 has just come out and using it here would make this config as future proof as possible. Is Python 3.7 chosen here for specific reasons?

.circleci/config.yml
```diff
@@ -1,63 +1,91 @@
...
+        #auth:
+        #  username: $DOCKERHUB_USERNAME
+        #  password: $DOCKERHUB_PASSWORD
```
jacksonllee, line 19: I think we can safely remove these lines. I don't envision the need for passing in Docker Hub credentials for this project.

.circleci/config.yml
```diff
@@ -1,63 +1,91 @@
...
+
+  build-python:
+    parameters:
+      python-version:
+        type: string
+    docker:
+      - image: cimg/python:<< parameters.python-version >>
+        #auth:
+        #  username: $DOCKERHUB_USERNAME
+        #  password: $DOCKERHUB_PASSWORD
```
jacksonllee, line 41: Same comment as above for removing these commented-out lines.

.circleci/config.yml
```diff
@@ -1,63 +1,91 @@
...
       - run:
-          name: Check styling consistency with black
-          working_directory: ~/
-          # Avoid being able to import wikipron by relative import.
-          # Test code by importing the *installed* wikipron in site-packages.
-          command: black --line-length=79 --check project/setup.py project/wikipron project/tests project/data
+          name: Create empty pyenv file
+          command: |
+              mkdir -p /home/circleci/.pyenv/versions/
+              touch /home/circleci/.pyenv/version
```
jacksonllee, line 48: Why do we need this `run` step for creating the `/home/circleci/.pyenv/versions/` directory and the `/home/circleci/.pyenv/version` file? Is it possible to remove this step, since I thought we're already running the later steps in an image specified above by `cimg/python:3.x`?

.circleci/config.yml
```diff
@@ -1,63 +1,91 @@
...
+          pre-install-steps:
+              - run:
+                  name: Build source distribution and install package from it
+                  command: |
+                      pip install --progress-bar off --upgrade pip setuptools
+                      python setup.py sdist
+                      pip install dist/*.tar.gz
+      - run:
+          name: Show installed Python packages
+          command: pip3 list -v
```
jacksonllee, line 60: `pip` (used in the previous step) instead of `pip3` for consistency?

.circleci/config.yml
```diff
@@ -1,63 +1,91 @@
...
-  build-python-3.8:
-    <<: *template
-    docker:
-      - image: python:3.8
-  build-python-3.9:
-    <<: *template
-    docker:
-      - image: python:3.9
+          command: |
+              sudo chown circleci:circleci /home
```
jacksonllee, line 67: Why do we need to run `chown` on `/home` here? A comment just above would help.

.circleci/config.yml
```diff
@@ -1,63 +1,91 @@
...
+              pytest -vv project/tests --junitxml /tmp/testxml/report.xml
+      - store_test_results:
+          path: /tmp/testxml/
```
jacksonllee, line 70: I notice that you've added `--junitxml /tmp/testxml/report.xml` to the pytest command above to generate a test report. Does this `store_test_results` step put the report somewhere? How do we access it?

jacksonllee (commented):
Thank you @platipo for this contribution! The CircleCI orbs are new to me. I think I see how they can be useful, as this pull request has shown. I like how the checks for flake8/black/etc. common to all Python versions are factored out and required to pass prior to running the test suite. Cached environments for less build time are also nice.

As for the changelog, would you mind adding a new entry like `Improved CircleCI workflow with orbs. (\#249)` to `CHANGELOG.md`?

Is there a label I should add to this pull request for the purpose of Orbtoberfest? Let me know and I'll do it.

jacksonllee:
Heads-up that #247 adding a mypy step to the CircleCi config has just been merged.

.circleci/config.yml
platipo: Starting from 1st November, CircleCI will limit the unauthenticated Docker pulls, you can read more [here](https://circleci.com/docs/2.0/private-images/).

.circleci/config.yml
platipo: I modified the command to `pip3` because in a latter image `pip` showed both Python 2 and 3 packages. Do you want me to remove also the `-v` flag that I used for debugging the package location? This is an [output sample](https://app.circleci.com/pipelines/github/kylebgorman/wikipron/558/workflows/c28a52d8-0b7c-40bf-95a7-6ce66168d0d5/jobs/1630/parallel-runs/0/steps/0-107)

.circleci/config.yml
platipo: I'll mark this as resolved so to continue the conversation on authenticated Docker images only in one place.

.circleci/config.yml
platipo: I added the `--junitxml /tmp/testxml/report.xml` and `store_test_results` because it is needed to split the test for the parallelization; you can access the results clicking on a build job and then selecting the test tab: [for example](https://app.circleci.com/pipelines/github/kylebgorman/wikipron/558/workflows/c28a52d8-0b7c-40bf-95a7-6ce66168d0d5/jobs/1630/tests)

platipo: It it also useful to read failure messages cleanly formatted

platipo:
Thank you @jacksonllee, I rebased the branch on master and added the mypy step; this is still a WIP but I'm adding the finishing touches :rocket:

.circleci/config.yml
platipo: I noticed that the [first test are failing](https://app.circleci.com/pipelines/github/kylebgorman/wikipron/560/workflows/997720a3-3515-4d7a-a01c-20604a8bbc5b/jobs/1642/tests) because they are trying to create a directory in `/home` which is owned by `root`.
I chose a quick and dirty fix to make tests pass; of course the correct fix is to move the test directory to either a unique temporary folder via `tempfle` stdlib or a fixed directory in a user owned path like `~/tsv`

.circleci/config.yml
platipo: In `python:3.x` images it isn't present `/home/circleci/.pyenv/versions/` and this blocked cache creation. In the a4189f9 commit I tried to remove this step and in `cimg/python:3.x` it seems that the directory is present and cache works as intended.

.circleci/config.yml
platipo: No, I had no specific reason in mind; I just chose a chose a version I knew it worked.

platipo:
As long as you don't want the workflow with separate parallel tests, I think I have finished if there are no other fix to do.

If tests became longer and longer the new workflow could become:

proposed workflow

In my opinion it is not worth it because there is a context switch to take into account from the build job to the test job for workspace save, but we can try...

platipo:
> Is there a label I should add to this pull request for the purpose of Orbtoberfest? Let me know and I'll do it.

Yeah, I need the `Orbtoberfest` label. Thank you!

.circleci/config.yml
jacksonllee: I see -- thanks for flagging this! In that case, I agree it's good to leave these commented-out lines here as a reminder to ourselves.

.circleci/config.yml
jacksonllee: Oh yes, we do have tests with file I/O like that. Your quick fix here seems reasonable for now.

Note to self: I did [mention](https://github.com/kylebgorman/wikipron/pull/226) using `tempfile` when the relevant tests were being checked in. This issue will be fixed when we get to refactoring the relevant code later.

.circleci/config.yml
jacksonllee: Interesting -- I must admit that I've been looking at test failures just by scrolling through the log on the "Steps" tab.

.circleci/config.yml
```diff
@@ -1,67 +1,91 @@
...
+          command-name: flake8
+          command-run: flake8 project/setup.py project/wikipron project/tests
+      - pre-build:
+          name: black
+          command-name: black
+          command-run: black --line-length=79 --check project/setup.py project/wikipron project/tests project/data
+      - pre-build:
+          name: mypy
+          command-name: mypy
+          command-run: mypy --ignore-missing-imports project/wikipron project/tests project/data
```
jacksonllee, line 82: Thank you for adding the mypy pre-build as well!

.circleci/config.yml
```diff
@@ -1,67 +1,91 @@
...
+                      python setup.py sdist
+                      pip install dist/*.tar.gz
       - run:
-          name: Check styling consistency with black
-          working_directory: ~/
-          # Avoid being able to import wikipron by relative import.
-          # Test code by importing the *installed* wikipron in site-packages.
-          command: black --line-length=79 --check project/setup.py project/wikipron project/tests project/data
+          name: Show installed Python packages
+          command: pip list -v
```
jacksonllee, line 55: The `-v` flag to be more verbose for where things are is nice. Thanks for adding it!

jacksonllee (approved):
LGTM! Thank you very much @platipo for showing us the CircleCI orb magic and patiently walking me through the details!

I went ahead to fix the merge conflict for `CHANGELOG.md` to save another round of back-and-forth and not take up more of your time. Thank you again!

PR #250, opened 2020-10-25
split tildes; resort

jimregan:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,8 @@ Unreleased
 -   Added Shan (`shn`) with custom extraction. (\#229)
 -   Added Northern Kurdish (`kmr`). (\#243)
 -   Added a script to facilitate the creation of a `.phones` file. (\#246)
+-   Split multiple pronunciations joined by tilde in `eng_us_phonetic`
+
 
 #### Changed
 
--- a/data/tsv/eng_us_phonetic.tsv
+++ b/data/tsv/eng_us_phonetic.tsv
@@ -282,11 +282,9 @@ certification	s ɚ ɾ ɪ f ɪ k e ɪ ʃ ə n
 chance	t ʃʰ a n s
 chance	t ʃʰ e ə n s
 chance	t ʃʰ æ n s
-chance	t ʃʰ æ n s ~ t ʃʰ a n s
 chance	t ʃʰ ɐː n s
 chance	t ʃʰ ɑː n s
 chance	t ʃʰ ɛ ə n s
-chance	t ʃʰ ɛ ə n s ~ t ʃʰ e ə n s
 chaparral	ʃ æ p ɚ ɹ æ l
 chaparral	ʃ æ p ɚ ɹ ɛ l
 char	t ʃ ɑ ɹ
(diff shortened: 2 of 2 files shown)


kylebgorman (approved):
Thanks for this fix, but this will not persist across the next scrape. We usually fix these issues by modifying the data `upstream` (that is, we edit the Wiktionary page for _Ireland_ or _chance_). However, I'm happy to accept the fix all the same, and I'll make the upstream edits. LGTM.

jimregan:
@kylebgorman I think it might be better to split while scraping: tilde for 'or' is a common convention, and it turns up in a couple of the other files:

```
    106 per_phonemic.tsv
     30 dsb_phonetic.tsv
      3 tuk_phonemic.tsv
      3 eng_uk_phonetic.tsv
      2 nds_phonemic.tsv
      2 cop_phonemic.tsv
      1 swe_phonetic.tsv
      1 mon_cyrl_phonetic.tsv
      1 kaz_phonetic.tsv
      1 ice_phonemic.tsv
      1 eng_us_phonemic.tsv
      1 eng_uk_phonemic.tsv
      1 dut_phonetic.tsv
```

I'm not familiar with Persian, but I am somewhat familiar with Lower Sorbian (dsb), where they're all splits. The [Icelandic](https://en.wiktionary.org/wiki/Elli#Icelandic) seems to come from "(when derived from Erlendur as pronounced /ˈɛtl(ɛ~ɪ)ntʏr/", and I'm not sure how to fix that in Wiktionary.

Python is not my go-to language, but I imagine it would look something like:
```
        for word, pron in config.extract_word_pron(title, request, config):
            # Pronunciation processing is done in NFD-space;
            # we convert back to NFC aftewards.
            normalized_pron = unicodedata.normalize("NFC", pron)
            # 'cast' is required 'normalize' doesn't return a 'Pron'
            if not '~' in normalized_pron:
                yield word, cast(Pron, normalized_pron)
            else:
                for partial in normalized_pron.split('~'):
                    yield word, cast(Pron, partial.strip())
```

kylebgorman:
You raise a good point. I would probably prefer to just flag and reject
though, I think? They're clearly not to Wiktionary style since they're used
so sparsely, and all the pronunciation templates support multiple entries.

In the past (though I don't see any right now) we saw tildes separates
partial pronunciations, and I think decoding the extension of an implicit
regular expression here is a bridge too far.

@lfashby @jacksonlee any thoughts?

lfashby:
Flag and reject would certainly be the most cautious solution. The most consistent solution would be to modify `_skip_pron` by adding:
```
if “~” in pron:
	return True
```
so it matches our previous treatment of `-` https://github.com/kylebgorman/wikipron/blob/551c97f555df6920ed8ac3798e5d957c9a44095e/wikipron/extract/core.py#L16-L18 but that doesn’t seem smart. I’m not sure why we don’t flag prons containing `-`.

jacksonllee:
The tilde has been a known issue, but I'm having trouble recalling why we didn't decide to do something about it within the WikiPron package here. Given the apparently small number of occurrences, I agree that @lfashby's suggestion of adding `if "~" in pron:` to check and reject pronunciations seems reasonable.

PR #251, opened 2020-10-25
Small path changes on the data side, rework of test_scrape.py

lfashby:

Quick rework of `scrape.py` and `test_scrape.py` so that we can use `tempfile`. I believe this was requested in #249


data/src/codes.py
```diff
@@ -34,12 +35,15 @@
 
 LANGUAGES_PATH = "languages.json"
 UNMATCHED_LANGUAGES_PATH = "unmatched_languages.json"
-README_PATH = "../README.md"
-LANGUAGES_SUMMARY_PATH = "../languages_summary.tsv"
 LOGGING_PATH = "scraping.log"
 ISO_639_1_PATH = "iso639_1-to-iso639_2.json"
 ISO_639_2_PATH = "iso639_2.json"
 URL = "https://en.wiktionary.org/w/api.php"
+DATA_DIRECTORY = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
```
kylebgorman, line 42: Is th iterated os.path.dirname necessary? I would have guessed naively that it'd be idempotent. Maybe you can say what this is computed in a comment if it is?

kylebgorman (approved):
LGTM. One candidate nit.

data/src/codes.py
lfashby: `os.path.dirname(os.path.realpath(__file__))` takes you to the `src` directory, the `os.path.dirname()` of that is the `data` directory. I guess there are a lot of ways to do this but I'm following the conventions of: https://github.com/kylebgorman/wikipron/blob/13be94162a42ff776dfa89655eb92b5613cfd53c/tests/test_wikipron/test_version.py#L7-L9

kylebgorman:
Okay, n/m, go ahead.

jacksonllee:
Sorry to be late to the party. If this PR successfully refactors `_build_scraping_config` to pass in paths more flexibly for allowing a temp directory in the test suite, then this should presumably avoid the need of altering the ownership of `/home` in the build container job [here](https://github.com/kylebgorman/wikipron/blob/551c97f555df6920ed8ac3798e5d957c9a44095e/.circleci/config.yml#L62) (a line that I would recommend removing).

PR #254, opened 2020-10-26
Style fixes for list_phones

kylebgorman:

### DIFF
--- a/data/phones/list_phones.py
+++ b/data/phones/list_phones.py
@@ -94,9 +94,7 @@ def main(args: argparse.Namespace):
         phone_to_examples.items(), key=lambda x: len(x[1]), reverse=True
     ):
         print(
-            f"{phone:<5}"
-            "\t# "
-            f"{len(examples):<10}"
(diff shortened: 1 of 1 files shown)


data/phones/list_phones.py
```diff
@@ -97,17 +95,16 @@ def main(args: argparse.Namespace):
     ):
         print(
             f"{phone:<5}"
-            " # "
+            "\t# "
             f"{len(examples):<10}"
```
jacksonllee, line 99: I was using the formatting like `:<5` and `:<10` to achieve what a tab character would do but with a bit more control. We should probably do one or the other, but not both?

kylebgorman: The `<5` thing isn't working for me (it seems to be counting in Unicode codepoints or bytes, not in number of spaces), hence the tab. But the `<10` bit is still wise IMO. See next commit.

jacksonllee: Makes sense -- LGTM.

Issue #141, opened 2020-04-16
Use positive, not negative flags


kylebgorman:
Flags should be renamed to be positive statements rather than negative ones, if possible.

kylebgorman:
Closed in #255.

PR #255, opened 2020-10-28
Negative flags are renamed to positive statements (#141)

yeonju123:
This pull request is for #141 
Negative flags in `cli.py` are renamed to positive statements. In order to accommodate this change,` Wikipron/config.py` and `tests/test_wikipron/test_config.py` are also edited accordingly.

### DIFF
--- a/wikipron/cli.py
+++ b/wikipron/cli.py
@@ -23,14 +23,16 @@ def _get_cli_args(args: List[str]) -> argparse.Namespace:
         ),
     )
     parser.add_argument(
-        "--no-stress",
-        action="store_true",
-        help="Remove stress marks in pronunciations.",
+        "--stress",
+        default=True,
+        help="By default, stress marks are included in pronunciations."
+        "To remove them, set this flag to False.",
     )
     parser.add_argument(
-        "--no-syllable-boundaries",
-        action="store_true",
-        help="Remove syllable boundary marks in pronunciations.",
+        "--syllable-boundaries",
+        default=True,
+        help="By default, syllable boundary marks are included."
+        "To Remove them from pronunciations, set this flag to False.",
     )
     parser.add_argument(
         "--dialect",
@@ -63,15 +65,15 @@ def _get_cli_args(args: List[str]) -> argparse.Namespace:
         ),
     )
     parser.add_argument(
-        "--no-segment",
-        action="store_true",
+        "--segment",
+        default=True,
         help=(
             "By default, the IPA pronunciation is segmented by whitespace and,"
             "to the extent possible, with a diacritic (either combining "
             "or modifier) immediately following the parent symbol. "
             'For example, "kʰæt" is segmented as "kʰ æ t", with kʰ '
             "conveniently segmented as an aspirated k for modeling purposes. "
-            "To disable such IPA segmentation, apply this flag."
+            "To disable such IPA segmentation, set this flag to False."
         ),
     )
     parser.add_argument(
@@ -85,9 +87,12 @@ def _get_cli_args(args: List[str]) -> argparse.Namespace:
         help="Skip spaces in the transcription",
     )
     parser.add_argument(
-        "--no-tone",
-        action="store_true",
-        help="Remove tones in the transcription.",
+        "--tone",
+        default=True,
+        help=(
+            "By default, tones are included in pronunciations."
+            "To remove them, set this flag to False."
+        ),
     )
     return parser.parse_args(args)
 
--- a/wikipron/cli.py
+++ b/wikipron/cli.py
@@ -24,15 +24,29 @@ def _get_cli_args(args: List[str]) -> argparse.Namespace:
     )
     parser.add_argument(
         "--stress",
+        action="store_true",
+        dest="stress",
         default=True,
-        help="By default, stress marks are included in pronunciations."
-        "To remove them, set this flag to False.",
+        help="Include stress marks in pronunciations.",
+    )
+    parser.add_argument(
+        "--no-stress",
+        action="store_false",
+        dest="stress",
+        help="Remove stress marks from pronunciations.",
     )
     parser.add_argument(
         "--syllable-boundaries",
+        action="store_true",
+        dest="syllable_boundaries",
         default=True,
-        help="By default, syllable boundary marks are included."
-        "To Remove them from pronunciations, set this flag to False.",
+        help="Include syllable boundary marks in pronunciations.",
+    )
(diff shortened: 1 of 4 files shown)


kylebgorman (commented):
Hi Yeonju, this looks great. A few minor things:

1. Could you update the CHANGELOG, commit and push that change, nd then click the checkbox on the PR description to indicate that you did so?
2. We discussed earlier using [metavar](https://docs.python.org/3/library/argparse.html#metavar) earlier to ease migration. Would you have time to do that? 

Re: #2, I think it would look a bit like this:

```python
parser.add_argument(
   "--stress', metavar="stress", action="store_true", ...
)
parser.add_argument(
   "--no-stress", metavar="stress", action="store_false", ...
)
```

for each of these fields. That way no users have to change their code. What do you think?

@jacksonllee if you have a second, would appreciate your thoughts on that proposal...

yeonju123:
Ah, I agree with you. For users who are already familiar with neg flags,
that will be much better:)

yeonju123:
I will edit it as @kylebgorman  suggested in another branch, and if @jacksonllee agrees, I will create a PR.

jacksonllee:
+1 for keeping the neg flags for backward compatibility. (I'd assigned this PR to myself as my code review reminder to myself. But now it sounds like Yeonju is going to push more changes here. So I'm de-assigning myself. Please feel free to tag me for code review again later.)

kylebgorman:
SGTM.

yeonju123:
I have two issues I am facing with. 

Metavar is not compatible with “store_true” and “store_false”. Instead, I tried to use dest, which Kyle suggested initially in our last conversation. 
As our number of arguments in cli.py has increased, I have to add the added positive flags in our config.py as well. Otherwise, it yields an assertion error in test_cli.py when I run pytest.
Would it be a desirable way? Or is there another way to get around this error?

Moreover, we used to have default values for the relevant arguments (stress, syllable_boundaries, tones, segment,). As we include positive flags, it seems to me that having default positive values for stress, syllable_boundaries, tones, segment, may be redundant. 

If we want to maintain the way users are familiar with already, I can do something like..
parser.add_argument(
   "--stress', dest="stress", action="store_true", default = True...
)
parser.add_argument(
   “—no-stress', dest="stress", action="store_fault", default = False...
)

I am open to suggestions!

Yeonju

kylebgorman:
Oh right, `dest`, that's what I meant, sorry! Slight variation on what you said upthread:

```python
parser.add_argument(
   "--stress", action="store_true", default = True...
)
parser.add_argument(
   “—no-stress", dest="stress", action="store_false", default = False...
)
```

note that I don't need to specify `dest` for `--stress`.

kylebgorman:
I actually don't know what's necessary to help with the assertion error in the test, though. Could you post that here?

yeonju123:
Following is the error that I got :)


def test_cli_args_match_config_args():
        config_args = inspect.getfullargspec(Config.__init__)
        cli_args = _get_cli_args(["eng"])
>       assert cli_args.__dict__ == {
            **config_args.kwonlydefaults,
            "key": "eng",
        }, "CLI and Config.__init__ must have the same args and their
defaults."
E       AssertionError: CLI and Config.__init__ must have the same args and
their defaults.
E       assert {'casefold': ...': 'eng', ...} == {'casefold': ...': 'eng',
...}
E         Omitting 10 identical items, use -vv to show
E         Left contains 1 more item:
E         {'syllable-boundaries': True}
E         Right contains 1 more item:
E         {'syllable_boundaries': True}
E         Use -v to get the full diff

tests\test_wikipron\test_cli.py:34: AssertionError

kylebgorman:
Don't know but this looks like there's an inconsistency between
"syllable_boundaries" and "syllable-boundaries". I think we'd prefer the
latter, tis a bit more standard.

yeonju123:
Ah, dest does not read ‘-‘ as ‘_’ !
I changed it and it works now! I am sorry for the confusion xD
Please disregard what I said about changing Config.py.

wikipron/cli.py
```diff
@@ -24,15 +24,29 @@ def _get_cli_args(args: List[str]) -> argparse.Namespace:
     )
     parser.add_argument(
         "--stress",
+        action="store_true",
+        dest="stress",
```
kylebgorman, line 28: This line is redundant I believe.

wikipron/cli.py
```diff
@@ -24,15 +24,29 @@ def _get_cli_args(args: List[str]) -> argparse.Namespace:
...
+    parser.add_argument(
+        "--no-stress",
+        action="store_false",
+        dest="stress",
+        help="Remove stress marks from pronunciations.",
     )
     parser.add_argument(
         "--syllable-boundaries",
+        action="store_true",
+        dest="syllable_boundaries",
```
kylebgorman, line 41: I think this one is also redundant because it maps `--syllable-boundaries` onto `args.syllable_boundaries`?

wikipron/cli.py
```diff
@@ -66,14 +80,24 @@ def _get_cli_args(args: List[str]) -> argparse.Namespace:
     )
     parser.add_argument(
         "--segment",
+        action="store_true",
+        dest="segment",
```
kylebgorman, line 84: Once again I think this is redundant (but harmless).

wikipron/cli.py
```diff
@@ -88,11 +112,16 @@ def _get_cli_args(args: List[str]) -> argparse.Namespace:
     )
     parser.add_argument(
         "--tone",
+        action="store_true",
+        dest="tone",
```
kylebgorman, line 116: Final one: think this is redundant.

kylebgorman (commented):
Okay, I think a few of these "dest" args are redundant, but that's the last thing that needs attention.

lfashby:
Hi all, are we not interested in renaming the `no_skip_spaces_pron` and `no_skip_spaces_word` flags?

kylebgorman (approved):
LGTM. Thanks Yeonju!

kylebgorman:
Oh, Lucas brings up a good point. I missed those.

yeonju123:
Oh I assumed we were doing it only for "linguistic" features. I will add them as well!

lfashby:
Since I came up with those terrible names perhaps I should have to fix them? Even though this is backward compatible, after we merge this in I'm thinking we should switch over all the previous flags on the `data` side to the shiny new positive flags.

kylebgorman:
Okay, I'll merge.

PR #267, opened 2020-11-14
[bul] Adds phone list, filtered Bulgarian data

### LINKED ISSUE #264: [bul] Create .phones list for Bulgarian
As captioned.

ajmalanoski:

Closes #264 

Some phonemes in `bul_phonemic.phones` did not actually appear in the data. One of these phonemes was /fʲ/, and seems to be an accidental gap – both of my sources give it as a full (rather than marginal phoneme) of Bulgarian (so long you analyze palatalized consonants as separate phonemes, rather than /Cj/ clusters; this is the analysis assumed in Wiktionary transcriptions for Bulgarian). The other cases involved affricates that were not "noticed" by `list_phones.py` because they weren't transcribed with tie bars. I placed these phonemes at the end of the file.


kylebgorman:
Wow this data is a mess. Thanks for this though.

kylebgorman (approved):
LGTM.

Issue #253, opened 2020-10-26
Restart command with partial extracted data 


xburelle:
I have use the command line command and, I think, I reached a limit from the server because I was disconnecting and this finished the command.
So it can be useful to continue from a word because data are extracted alphabetically.

jacksonllee:
Thank you for reporting this. Do you by any chance have any logging and/or errors from the terminal that you could copy and share here? Also, how far did the run go (e.g., how many word-pronunciation pairs had been retrieved, if this is known) before you ran into this issue?

kylebgorman:
We also provide [pre-scraped data](https://github.com/kylebgorman/wikipron/tree/master/data) for about 175 languages, and update this periodically (once or twice a year).

xburelle:
Unfortunately, I don't have logs anymore. But I remember it was a "server can't be reached" error. May be my local network or, as I was using wikipron in english and in french at the same time, it can be a security from wiktionary servers. 
I have 32131 lines in english (UK) and 38386 in french, at the end (before the issue appears for both).
Thanks for helping

kylebgorman:
Sometimes the server just hangs up or times out, and this may be one of them. When we do our biannual (twice a year, not once every two years) scrape, we run them in serial because early on we inferred that it was hanging up if we made too many connections.

lfashby:
I’ve been thinking about this problem and something came up that might be the first step toward getting a restart command to work. I can’t recall if we already considered and dismissed trying this months ago...
It is possible to start a scrape from any entry within a language given that you know the `sortkey` for that entry. For example, if I run a scrape of Basque with these parameters:
```
    requests_params = {
        "action": "query",
        "format": "json",
        "list": "categorymembers",
        "cmtitle": category,
        "cmlimit": "500",
        "cmprop": "ids|title|timestamp|sortkey",
        "cmsort": "sortkey", 
        "cmstarthexsortkey": "4e454e44494e0a4e454e44494e"
    }
```
The scrape will start on the word 'nendin' and proceed as normal from there — grabbing the 500 category members after ‘nendin’. Thus if we store the `sortkey` of the most recently scrapped entry (which we have access to given the change to `cmprop` above), and catch whatever error occurs when the connection drops, we ought to be able to automatically restart the scrape from the last `sortkey`. Granted this description is a little bit short on the specifics of how to implement this restart, but I wonder if you think something like this might be possible?

kylebgorman:
I think this would be possible but a lot of work to get right.

[...]

PR #277, opened 2020-11-30
[yue] Handle Cantonese for scraping

jacksonllee:
This PR adds the Cantonese-specific extraction functions; these changes are very similar to #124 for Mandarin Chinese. 

This is part of #57, but does not fully resolve it. For reasons I have yet to understand better, I wasn't able to scrape for the Cantonese data. I triggered the scraping run with `./data/src/scrape.py --restriction=yue`, but kept getting `Exception detected while scraping: 'yue', '/Users/jacksonlee/repos/wikipron/data/tsv/yue_phonemic.tsv', ''` several times from the log after hours of scraping. It looks like the TSV file `data/tsv/yue_phonemic.tsv` was being filled, but only up to about a thousand or so entries, and then the scraping restarted by wiping this TSV file clean and showing that log line. Not entirely sure yet, but this is possibly related to #253 regarding rate limit etc.

These changes work for Cantonese as I've verified with the command `wikipron yue`. I thought I should get them checked in first, even though the TSV data isn't available yet.

### DIFF
--- a/wikipron/extract/yue.py
+++ b/wikipron/extract/yue.py
@@ -13,6 +13,10 @@ if typing.TYPE_CHECKING:
     from wikipron.typing import Iterator, Word, Pron, WordPronPair
 
 
+# The last bit of //ul//li[small[i[a[@title="w:Standard Cantonese"]]]]
+# is needed because some Cantonese entries have pronunciation info for
+# multiple dialects (Standard Cantonese, Taishanese, etc.), and we need
+# to specifically target Standard Cantonese.
 _PRON_XPATH_TEMPLATE = """
     //div[@class="vsHide"]
         //ul
--- a/wikipron/extract/yue.py
+++ b/wikipron/extract/yue.py
@@ -1,0 +1,36 @@
+"""Word and pron extraction for Cantonese."""
+
+import itertools
+import typing
+
+import requests_html
+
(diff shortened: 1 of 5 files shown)


wikipron/extract/yue.py
```diff
@@ -0,0 +1,40 @@
...
+
+if typing.TYPE_CHECKING:
+    from wikipron.config import Config
+    from wikipron.typing import Iterator, Word, Pron, WordPronPair
+
+
+# The last bit of //ul//li[small[i[a[@title="w:Standard Cantonese"]]]]
+# is needed because some Cantonese entries have pronunciation info for
+# multiple dialects (Standard Cantonese, Taishanese, etc.), and we need
+# to specifically target Standard Cantonese.
```
jacksonllee, line 19: Here's a visual for what this comment alludes to (for the character [一](https://en.wiktionary.org/wiki/%E4%B8%80#Pronunciation) "one"):

Screen Shot 2020-11-30 at 7 59 45 AM

kylebgorman:
Hi Jackson, reviewing now. So, that pattern is exactly what `scrape.py` does---it restarts and overwrites if it catches an exception (usually a rate limiting from the server).

I'm fine with merging this without the actual TSV files, since we can work on that later.

kylebgorman (approved):
LGTM.

PR #286, opened 2020-12-15
[deu] Split into Swiss German and Germany German. [nep] removes a duplicate from `languages.json`

yeonju123:

`ger` is split into Swiss German and Germany German.
`Nep` duplicate is removed from `languages.json`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,8 @@ Unreleased
 
 #### Changed
 
+-   Split `ger` into Germany German and Swiss German. (\#286)
+-   Remove a duplicate of `nep` from `languages.json`. (\#286)
 -   Improved printing in the README table. (\#145)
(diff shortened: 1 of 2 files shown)


kylebgorman:
Hi Yeonju, this looks good to me.

Can you run the scrape for these languages in your branch to see how this affects the counts? I think you'd follow the instructions here:

https://github.com/kylebgorman/wikipron#repository

then you'd navigate to `data/src` and run `./scrape --restriction=ger && ./postprocess && ./generate_summary`.

yeonju123:
Ok! I will work on that :)

Yeonju

kylebgorman:
Just in the interest of getting rid of old bugs, etc., I looked into this and it seems to now be out of date. We have support for `gsw`, and so there are there are only three words which would be affected by this,and two of them are MWEs. Closing for now.

PR #303, opened 2020-12-31
Rescrapes Armenian.

### LINKED ISSUE #301: [arm] missing Western Armenian entries
Hi again

I think your Western Armenian [dataset](https://github.com/kylebgorman/wikipron/blob/master/data/tsv/arm_w_phonetic.tsv) is missing a lot of words (potentially almost all of them). For example, this [word](https://en.m.wiktionary.org/wiki/%D5%A3%D6%80%D5%BA%D5%A1%D5%B6) has a Western pronunciation on Wiktionary but it's not on your dataset. 

I suspect the problem is that most of the Armenian transcriptions are all generated via a [script](https://en.wiktionary.org/wiki/Module:hy-pronunciation) (explained [here](https://github.com/kylebgorman/wikipron/issues/298)) whereby:
1. A wiktionary user writes a transcription using the Armenian script
2. The script then converts the transcription into IPA for Eastern and Western Armenian.
The users use this script because for a large chunk of the lexicon, it's predictable to go from the Eastern pronunciation towards the Western one via a series of sound changes.

It looks like your code just stops at capturing the Eastern transcription and doesn't then capture the Western transcription.

kylebgorman:
Closes #301.


lfashby (approved):
LGTM, thanks for taking care of this.

PR #330, opened 2021-01-23
[nob] Adds Østnorsk (Bokmål) phones and updated TSV data

### LINKED ISSUE #328: [nob] Create Urban East Norwegian (Bokmål) phonelist
As captioned

ajmalanoski:

Closes #328


ajmalanoski:
The data were transcribed really inconsistently in terms of what symbols were used for which phonemes, as noted in the `.phones` file. The "less common transcriptions" listed in the file nevertheless occur quite often (otherwise I wouldn't have bothered including them in the list). Given the small amount of data overall, I probably wouldn't recommend Bokmål for the shared task unless the data gets cleaned up.

data/phones/nob_phonemic.phones
```diff
@@ -0,0 +1,67 @@
...
+ʉː
+ə
+u
+uː
+o  # Less common transcription of /ɔ/.
+oː
+ɔ
+ɑ
+ɑː
+# DIPTHONGS
```
kylebgorman, line 57: typo: should be DIPHTHONGS (two <h>'s).

data/phones/nob_phonemic.phones
```diff
@@ -0,0 +1,67 @@
...
+# Norwegian has several diphthongs, which were not transcribed with tie bars.
+# As such, the following do not appear in the data.
+œ͡ʏ
+æ͡ɪ
+æ͡ʉ
+### PITCH ACCENT ###
+# Norwegian has pitch accent. However, it is either very rare, or is not 
+# transcribed very often on Wiktionary. Additionally, it is transcribed with
+# non-standard IPA (<¹> and <²>), so it is not included here. The pitch accent
+# is not reflected in the orthography.
```
kylebgorman, line 67: Re: this I thought that this standard does not have a pitch accent but Nynorsk does...but please feel free to check me. I know next to nothing about this language.

data/phones/nob_phonemic.phones
```diff
@@ -0,0 +1,67 @@
...
+ç
+j
+k
+ɡ
+ŋ
+h
+# Syllabic consonants
+n̩
+### VOWELS ###
+i  # Less common transcription of /ɪ/.
```
kylebgorman, line 31: Huh so it looks like they couldn't decide whether to transcribe quality differences for quantity differences...this problem comes up in the transcription of pretty much every Germanic language I know of. (In English it's pretty clear monophthong length is tense vs. lax nowadays; in German it's not obvious to me that closed-syllable monophthongs are really laxer but they are usually transcribed like that, and so on...)

kylebgorman (approved):
This LGTM though some discussion below.

data/phones/nob_phonemic.phones
ajmalanoski: I only know what Wikipedia told me, but it said that most varieties of Norwegian (including in Oslo, the speech of which Østnorsk is largely based on) have pitch accent.

data/phones/nob_phonemic.phones
ajmalanoski: Yep, that's exactly it. Although there wasn't a clear preference for one or the other overall, it seemed – for some pairs, they preferred length distinctions (_e.g._, /e/ was more common than /ɛ/), while for other they preferred tenseness (_e.g._, /ɪ/ was more common than /i/). But in a lot of cases, both styles were very common. For example, <ɪ> occurs 157 times, but <i> occurs 112 times – about 10% of all the entries!

data/phones/nob_phonemic.phones
ajmalanoski: Right you are! Fixed.

data/phones/nob_phonemic.phones
kylebgorman: I think this is something we could fix by bot.

Possibly a dumb question, but did you see any correlation with whether the syllable was closed or open?

ajmalanoski: I looked through the TSV file a bit to compare /ɪ/ and /i/, and I didn't really notice any patterns. Both seem to occur in the same contexts, as far as I can tell

kylebgorman:
Okay, then, I think we should look at Phoible (for instance) and then
recommend a bot-based standardization to Aryaman.

kylebgorman:
Checked---there's only one Phoible inventory (there are two actually, but
they're sourced from the same 1972 grammar), but both have tense/lax+length
distinctions, so I'd probably choose to standardize on that, all else held
equal.

ajmalanoski:
Makes sense to me.

PR #331, opened 2021-01-24
[arm] cleaned up armenian phones

### LINKED ISSUE #298: [arm] issues in the phones list
Wonderful resource. 

There are some errors in the [phone](https://github.com/kylebgorman/wikipron/blob/master/data/phones/arm_e_phonetic.phones) list for the Armenian dialects. A lot of the errors got cleaned up since you last scraped them.

The way that the Wiktionary contributors for Armenian work is that they do the following:

1. They take the orthographic form like գրպան (transliteration is `<grban>`, pronunciation is [gərbɑn])
2. They manually the orthographic and rewrite it with Armenian letters in order to apply any phonological rules like schwa insertion: գրպան to գըրպան <gərban>
3. They then use [this](https://en.wiktionary.org/wiki/Module:hy-pronunciation) script to convert from the rewritten form (2) into the IPA for Eastern and Western Armenian
4. This means that the Western entries are almost all redundant and automatically derived from the Eastern entries. A lot of words actually include both IPA entries together.

For [Eastern](https://github.com/kylebgorman/wikipron/blob/master/data/phones/arm_e_phonetic.phones) (EA)
* You're right that the `<ա>` grapheme is /ɑ/. The script automatically converts the rewritten grapheme to [ɑ]. Some people had manually written the pronunciation entries and used non-IPA symbols like [a]. But they've mostly gotten cleaned up. A lot got cleaned before August. Then I personally scraped Wiktionary with another python package, found some overlooked [a]'s, and I cleaned them up too.
* Same issue with  `<ո>` actually being [ɔ] but sometimes incorrectly written as [o].
* Armenian doesn't have phonemic geminates. What you see is that the orthography has a sequence of identical consonants like փթթել `<pttel>` [pəttel]. The script automatically converts a sequence of identical segments into a geminate with the [ː] diacritic. So if the rewritten form is փըթթել `<pəttel>`, then the outputted pronunciation is [pət:el]. If the rewritten form is փըթել `<pətel>`, then the automatic output is [pətel]. IMO, it would make more sense if the automatic transcription was with doubled segments [pəttel] because Armenian lacks phonemic length. So all the consonants have a possible geminated/lengthened/doubled form. It might just be an accidental gap in the Wiktionary data that you don't have geminate t͡s: and others.
* The back fricatives are in free variation between velar and uvular, with more tendency towards uvular. The automatic transcription uses χ, so any [x] that you see is just an error (which I think also got cleaned up in around summer).
* What's the issue with missing tie-bars?

    * TODO: for future cleanup, missing tie-bars on segments like:
  <ց> [tʃʰ]
   <չ> [tsʰ]

For [Western](https://github.com/kylebgorman/wikipron/blob/master/data/phones/arm_w_phonetic.phones) (WA)
* Any [r] that you see is an error. The orthography has two rhotic graphemes ր ռ that are both pronounced the same as a flap. The trill is just someone manually writing it. That counts as an error and I think most of them got cleaned up on the WIktionary entries.
* Yup you're right about the mergers.
    * TODO: supposedly the following mergers have occurred in this dialect:
  aspirated              > aspirated 
  voiced                 > aspirated
  voiceless unaspirated  > voiced
* The W dialect also developing a devoicing rule for clusters of orthographic voice+voiceless consonants. The automatic transcription covers that though.

jhdeov:
Cleans up the phone lists for `arm_e` and `arm_w`. Closes #298.

### DIFF
--- a/data/phones/arm_e_phonetic.phones
+++ b/data/phones/arm_e_phonetic.phones
@@ -9,11 +9,17 @@ i
 ɔ
 u
 #
-# Older entries might contain [o] or [e]. These have been fixed. You shouldn't find these phones anymore
+# Older entries might contain [o] or [e]. These have been fixed. You shouldn't find these
+# phones anymore.
 #
-# Long vowels: A sequence of identical heterosyllabic vowels is phonetically pronounced as one long vowel. Armenian does not have phonemic vowel length. But these are very rare. The only attested case on Wiktionary is the following
+# Long vowels: A sequence of identical heterosyllabic vowels is phonetically pronounced as 
+# one long vowel. Armenian does not have phonemic vowel length. But these are very rare. 
+# The only attested case on Wiktionary is the following.
 ɛː
-# The other types of long vowels are not found on Wiktionary as of Jan 2021. This is largely because long vowels are mostly in loanwords and borrowings, which Wiktionary underreports. At some point in the future, the Armenian users might add some entries which contain these sequences.
+# The other types of long vowels are not found on Wiktionary as of Jan 2021. This is 
+# largely because long vowels are mostly in loanwords and borrowings, which Wiktionary 
+# underreports. At some point in the future, the Armenian users might add some entries 
+# which contain these sequences.
 # ɑː əː iː ɔː uː
 #
 #
@@ -50,12 +56,14 @@ j
 r
 ɾ
 #
-# Past errors: some instances of [χ] were incorrectly transcribed as [x]. These were fixed.
-# Past errors: some affricates were missing a tie-bar. These were fixed.
+# Past errors: some instances of [χ] were incorrectly transcribed as [x]. 
+# Past errors: some affricates were missing a tie-bar. 
 #  * <ց> [tʃʰ] 
 #  * <չ> [tsʰ]
-#
-# Long consonants: A sequence of identical consonants are phonetically pronounced as long consonant. Armenian does not have phonemic consonant length or geminates
+# These were fixed.
+# 
+# Long consonants: A sequence of identical consonants are phonetically pronounced as 
+# long consonant. Armenian does not have phonemic consonant length or geminates.
 mː
 nː
 pʰː
@@ -78,6 +86,9 @@ zː
 lː
 ɾː
 #
-# As an accidental gap,  as of Jan 2021, there are no Wiktionary entries that have following long consonants:
+# As an accidental gap,  as of Jan 2021, there are no Wiktionary entries that have the
+# following long consonants:
 # t͡sː d͡zː d͡ʒː ʒː hː jː
-# This is just an accidental gap. The above long consonants can exist in Armenian. At some point in the future, the Armenian users might add some entries which contain these sequences
+# This is just an accidental gap. The above long consonants can exist in Armenian. 
+# At some point in the future, the Armenian users might add some entries which contain 
+# these sequences.
--- a/data/phones/arm_e_phonetic.phones
+++ b/data/phones/arm_e_phonetic.phones
@@ -1,66 +1,83 @@
 # Based on https://en.wikipedia.org/wiki/Armenian_language#Phonology
+# And based on the pronunciation script from Wiktionary: https://en.wiktionary.org/wiki/Module:hy-pronunciation
+#
 # Vowels.
 ɑ
 ɛ
 ə
 i
 ɔ
-o  # <ո> is really [ɔ].
-a  # <ա> is really [ɑ].
 u
+#
+# Older entries might contain [o] or [e]. These have been fixed. You shouldn't find these phones anymore
+#
+# Long vowels: A sequence of identical heterosyllabic vowels is phonetically pronounced as one long vowel. Armenian does not have phonemic vowel length. But these are very rare. The only attested case on Wiktionary is the following
+ɛː
+# The other types of long vowels are not found on Wiktionary as of Jan 2021. This is largely because long vowels are mostly in loanwords and borrowings, which Wiktionary underreports. At some point in the future, the Armenian users might add some entries which contain these sequences.
+# ɑː əː iː ɔː uː
+#
+#
 # Consonants.
 m
-mː
 n
-nː
 ŋ
 pʰ
-pʰː
 p
-pː
 b
-bː
 tʰ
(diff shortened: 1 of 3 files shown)


data/phones/arm_w_phonetic.phones
```diff
@@ -1,71 +1,79 @@
...
+l
+j
+ɾ
+#
+# Past errors: some instances of [χ] were incorrectly transcribed as [x]. These were fixed.
+# Past errors: some affricates were missing a tie-bar. These were fixed.
+#  * <ց> [tʃʰ] 
+#  * <չ> [tsʰ]
+#
+# Long consonants: A sequence of identical consonants are phonetically pronounced as long consonant. Armenian does not have phonemic consonant length or geminates. This is just an accidental gap. The above long consonants can exist in Armenian. At some point in the future, the Armenian users might add some entries which contain these sequences
```
kylebgorman, line 51: Thanks for these detailed comments: should prevent our future selves from making these mistakes in the future.

Minor nit: could you add trailing sentential punctuation to these comments? You might also want to hard-wrap them at 80 chars. There was an example above of this.

kylebgorman (commented):
This basically LGTM but I'm going to ask @ajmalanoski to take a closer look. I think we'll also want to re-run the scrape for this language to see how things are affected, but we can do that on our side so you don't have to bother.

If you could, please edit the CHANGELOG.md file at the top of the repository directory tree. You'll add a line to the section "Unreleased" > "Under `data/`". Then commit that change and click the little checkbox in the PR window here.

data/phones/arm_w_phonetic.phones
jhdeov: Sure. I tidied up. You should see the commit message "cleaned up armenian phones (with more tidying up)"

jhdeov: Rats, I placed it on the wrong spot on the changelog -- I fixed it.

data/phones/arm_e_phonetic.phones
```diff
@@ -1,66 +1,94 @@
...
+#
+# Long vowels: A sequence of identical heterosyllabic vowels is phonetically pronounced as 
+# one long vowel. Armenian does not have phonemic vowel length. But these are very rare. 
+# The only attested case on Wiktionary is the following.
+ɛː
+# The other types of long vowels are not found on Wiktionary as of Jan 2021. This is 
+# largely because long vowels are mostly in loanwords and borrowings, which Wiktionary 
+# underreports. At some point in the future, the Armenian users might add some entries 
+# which contain these sequences.
+# ɑː əː iː ɔː uː
```
ajmalanoski, line 23: If these are accidental gaps, could you please uncomment these and put them on their own lines? The same applies to the "missing" affricates, and the gaps in the file as well.

It's not a problem to have phones in the `.phones` file that don't currently appear on Wiktionary, whereas it will be a problem later on if someone adds a word with, say, /ɑː/, and we accidentally (and unknowingly) filter it out.

data/phones/arm_e_phonetic.phones
```diff
@@ -1,66 +1,94 @@
...
+ʁː
+lː
 ɾː
-# TODO: for future cleanup, missing tie-bars on segments like:
-# * <ց> [tʃʰ]
-# * <չ> [tsʰ]
+#
+# As an accidental gap,  as of Jan 2021, there are no Wiktionary entries that have the
+# following long consonants:
+# t͡sː d͡zː d͡ʒː ʒː hː jː
```
ajmalanoski, line 91: In other languages on Wiktionary, I have seen geminate affricates transcribed as /tts/ or /t:s/ instead of /t͡sː/ (to take one example). Also, if an affricate is transcribed without a tie bar, then our script treats it as a sequence of multiple segments, rather as one sound. For example, if /ts/ is transcribed as such, without a tie bar, then it will be segmented as the sequence /t/+/s/, rather than as a single affricate. Sometimes, this can make it look like affricates are missing from the data when they're actually present.

So my comment here is, have you checked that the missing long affricates aren't being transcribed in a "non-standard" way?

ajmalanoski (commented):
Hi @jhdeov, thanks for helping out. I left a couple comments: one with some suggested edits, and one with a question about the data. Otherwise, it looks good to me.

data/phones/arm_e_phonetic.phones
jhdeov: I fixed the comments

data/phones/arm_e_phonetic.phones
jhdeov: Yup. Many of the early entries were manually typed without using the tie-bar diagraph. I converted all those into entries that use the hy-pron script which automatically converts the affricate letters into the right affricate IPA segments with tie-bars.
For geminate affricates, they all look like [t͡sʰː]

kylebgorman:
@ajmalanoski is this ready, in your opinion? LGTM.

ajmalanoski:
It looks good to me, as well!

kylebgorman:
Thanks Hossep and congrats, you're now a contributor to WikiPron!

kylebgorman (approved):

ajmalanoski:
@jhdeov I'm running a scrape of Armenian right now. I've noticed that [ʏ] occurs very often in the West Armenian data (I didn't count, but there are dozens of instances). I just wanted to confirm than this phone is indeed _not_ found in Western Armenian.

jhdeov:
Rats. [ʏ] is an allophone of /ju,ju/ and the hy-pron [script](https://en.m.wiktionary.org/wiki/Module:hy-pronunciation) generates it. Apologies for missing that. The [ʏ] segment is only found in Western Armenian, not Eastern Armenian.

I reread the script and you'll find that apparently they also generate [œ] as a foreign phoneme in Western. I checked my scraped data and this segment is only found once.

ajmalanoski:
Okay, so if I understand correctly, the `.phones` file for Western Armenian should include [ʏ] and [œ]?

jhdeov:
@ajmalanoski  Correct, with a note like this

ʏ # allohpone that is automatically transcribed from word-medial յու /ju/
œ # rare and a marginal borrowed phoneme

ajmalanoski:
Okay, I can take care of the edits. Thanks!

PR #334, opened 2021-01-25
Partial scrape

lfashby:

Scrape of all languages except Cantonese, Russian and Chinese. There are a few things that came to my attention during this scrape and I'll leave some comments on the relevant issue tickets/investigate them shortly. 

[Kurdish](https://en.wiktionary.org/wiki/Category:Kurdish_language) (`kur`) no longer contains any entries. See the note on its Wiktionary page. I've therefore removed its TSVs.
[Opata](https://en.wiktionary.org/wiki/Category:Opata_language) (`opt`) no longer contains any entries, though I'm not sure why. I've removed its TSVs as well.

I've got a scrape of the remaining languages going, though even with the (experimental) restart functionality it will probably take a few days to finish. If the restart functionality works perhaps we should wait to do a release until after the scrape finishes, as the restart functionality does involve changes to the `wikipron` module.


kylebgorman:
Agree on holding off the release---that's a big feature.

kylebgorman (approved):
Wow, this is a ton of new languages. @jacksonllee's intuition that we wanted code we could run again and again was definitely the right one.

Issue #259, opened 2020-10-31
[nan] Support Min Nan


agutkin:
Supporting [Min Nan](https://en.wiktionary.org/wiki/Category:Min_Nan_terms_with_IPA_pronunciation) requires writing custom extractor.

kylebgorman:
Assigning this to this Sasha Gutkin character.

lfashby:
Based on the `nan` results from the last scrape (it ultimately had less than 100 entries and thus didn't make it through) we can scrape Min Nan, we just skip most entries because the headwords contain dashes. Since our [skip word logic](https://github.com/kylebgorman/wikipron/blob/master/wikipron/scrape.py#L52) occurs before our pron extraction we probably need to modify [`skip_word`](https://github.com/kylebgorman/wikipron/blob/master/wikipron/scrape.py#L27) to not skip words with dashes if  `config.key == 'nan'` (or something like that), rather than writing a custom extractor.

agutkin:
Can someone take this on? I don't have any cycles at the moment.

kylebgorman:
Oh that's very simple then, @lfashby can you take a stab at this as part of your larger process?

lfashby:
Sure, I can implement my crude solution and re-scrape. Perhaps at some point in the future we can come back to this if we think of a more elegant way of dealing with it/find other languages that have the same problem.

jacksonllee:
Just left [this comment](https://github.com/kylebgorman/wikipron/pull/340#discussion_r567214544) about Min Nan. Chinese languages are tough...

PR #340, opened 2021-01-29
Finishes scrape, adds restart command

### LINKED ISSUE #57: [yue] Cantonese support

### LINKED ISSUE #253: Restart command with partial extracted data

lfashby:

This adds Cantonese, Russian, and Chinese data. Also adds Min Nan data, which necessitated modifying `_skip_word()` in `wikipron/scrape.py` and updating `test_wikipron/test_scrape.py` slightly.

Finally this adds restart functionality to the wikipron module, which was used to scrape the data for the aforementioned languages. The addition of this restart functionality required some simplification of `src/scrape.py`, and modifications to `config.py` and `wikipron/scrape.py`.
This restart functionality assumes that we only ever experience timeouts/connection errors when trying to connect to a new headword (the get request in `_scrape_once()`) and therefore it grabs the `sortkey` associated with each headword before the get request, and then restarts the scrape from that `sortkey` should the get request fail. In this way, it never prints duplicate entries.

Based on my testing, this assumption of where we experience timeouts appears to be correct. I never experienced any connection loss in the middle of scraping a headword (say, while scraping the 2nd of 4 phonetic transcriptions for a headword) or from our get request that connects to the Wiktionary API. I suppose that doesn't mean either of those are impossible (though I think the former might be), but should we find evidence of them we can revise the restart functionality.

Closes #253 
Closes #57


lfashby:
Hi @jacksonllee, if you have some free time your feedback on the changes I've made to`config.py` and `wikipron/scrape.py` in order to implement the restart functionality would be most appreciated.

wikipron/scrape.py
```diff
@@ -1,11 +1,13 @@
 import re
 import unicodedata
-from typing import cast
+from typing import cast, Dict, Any
```
kylebgorman, line 3: Very minor note but could you sort these lexicographically?

wikipron/scrape.py
```diff
@@ -1,11 +1,13 @@
...
 
 import pkg_resources
 import requests
 import requests_html
 
+import time
```
kylebgorman, line 9: This is a built-in, not 3rd party, module so it should be on line 2.

Also there should be a blank line before the typing imports (but that's a "drive-by" issue that you didn't cause...it was already like that).

wikipron/scrape.py
```diff
@@ -24,15 +26,15 @@
...
+def _skip_word(word: str, skip_spaces: bool, language: str) -> bool:
     # Skips reconstructions.
     if word.startswith("*"):
         return True
     # Skips multiword examples.
     if skip_spaces and (" " in word or "\u00A0" in word):
         return True
     # Skips examples containing a dash.
-    if "-" in word:
+    if "-" in word and language != "Min Nan":
```
kylebgorman, line 37: Could you leave a TODO here? We ought to add this to the config setup someday, methinks.

wikipron/scrape.py
```diff
@@ -80,22 +83,36 @@ def scrape(config: Config) -> Iterator[WordPronPair]:
...
             "https://en.wiktionary.org/w/api.php?",
             params=requests_params,
             headers=HTTP_HEADERS,
         ).json()
-        yield from _scrape_once(data, config)
-        if "continue" not in data:
-            break
-        continue_code = data["continue"]["cmcontinue"]
-        requests_params.update({"cmcontinue": continue_code})
+        try:
```
kylebgorman, line 100: Nice, this is pretty straightforward actually!

kylebgorman (commented):
A few minor comments but other than that LGTM, this is great.

Once this is in are we ready to release?

kylebgorman:
Could you add "Closes #253" to the description here, Lucas?

kylebgorman:
I would love to hear from Jackson too before we do this.

lfashby:
I think I've addressed your comments suitably. What's the status of #259 given these changes? Should we leave it around to remind us to rework the dash stuff or file a separate issue?

I'd say we are good for a release after this. I'm also in favor of waiting for Jackson before merging this.

kylebgorman:
I would leave #259 open. We should just add that as a config param.

Sounds good, let's see if Jackson can take a look this weekend before we
submit. But then I'll do a release as soon as this is mergable.

>

jacksonllee:
I'm tied up at work at the moment, but can take a look tonight.

On making a new release, Kyle, you may want to take over #235 (I _think_ most if not all items listed there are done, but someone needs to check), for the version bump etc.

kylebgorman:
Will do re: #235.

kylebgorman:
@jacksonllee when I do a release should this be 1.2.0? I don't really feel comfortable making semantic versioning decisions like this...

wikipron/scrape.py
```diff
@@ -24,15 +26,17 @@
 }
 
 
-def _skip_word(word: str, skip_spaces: bool) -> bool:
+def _skip_word(word: str, skip_spaces: bool, language: str) -> bool:
```
jacksonllee, line 29: Revert the `_skip_word` changes if we aren't checking in Min Nan data?

jacksonllee (commented):
The updated restart code is great! For the scraped data, I would recommend dropping Min Nan (as well as the corresponding changes in `_skip_word`) in this pull request -- details in my comments.

jacksonllee:
> when I do a release should this be 1.2.0?

@kylebgorman There have been no API-breaking changes since version 1.1.0, correct? No breaking changes either for changes related to CLI flags, right? If that's the case, then version 1.2.0 is right.

kylebgorman:
Thanks for clarification Jackson.

I think Jackson's proposal to split out the Min Nan change is a good idea,
Lucas, if it's not too much trouble.

data/README.md
```diff
@@ -193,6 +194,7 @@
 | [TSV](tsv/okm_phonemic.tsv) | okm | Middle Korean (10th-16th cent.) | Middle Korean | False | Phonemic | 334 |
 | [TSV](tsv/gml_phonemic.tsv) | gml | Middle Low German | Middle Low German | True | Phonemic | 170 |
 | [TSV](tsv/wlm_phonemic.tsv) | wlm | Middle Welsh | Middle Welsh | True | Phonemic | 144 |
+| [TSV](tsv/nan_phonetic.tsv) | nan | Min Nan Chinese | Min Nan | True | Phonetic | 431 |
```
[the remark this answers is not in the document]
lfashby: Good point. Bit of an oversight on my part! I guess it is back to the drawing board on Min Nan and it looks like it'll be quite a challenge.

If we have to scrape through the Chinese category twice (phonetic/phonemic) for each of those Min Nan subdialects we'd probably add another 3-4 days to the big scrape runtime.

lfashby:
I removed all the Min Nan changes and updated the description to at last close the Cantonese support ticket.

jacksonllee (approved):
LGTM

kylebgorman (approved):
LGTM.

Issue #319, opened 2021-01-15
Automatic script inspection


lfashby:
Our current `data`-side method for handling languages that use multiple (orthographic) scripts is to manually specify in [`languages.json`](https://github.com/kylebgorman/wikipron/blob/master/data/src/languages.json) that multiple scripts will be present in TSVs for certain languages. [`split.py`](https://github.com/kylebgorman/wikipron/blob/master/data/src/split.py) then separates these scripts into different TSVs such that each TSV only contains one script.

The process of manually inspecting TSVs/Wiktionary to figure out if a language uses multiple scripts is so 2020 and ideally should be replaced by something that can automatically inspect each file in `data/tsv` and tell us whether more than one (Unicode) script is present and, if more than one script is present, what the Unicode script names are for each present script. For the time being, I imagine that the code running this automatic inspection should be placed in `tests/test_data`. If we manage to write something that can do this automatic inspection we can decide if it should be thoroughly integrated into the `data` side later.

This would be a cool (though I anticipate difficult to implement) feature. It’s a bit of a wonder that this technology doesn’t already exist, or perhaps it does and I haven’t found it. People have worked on things like it – see [this stackoverflow issue](https://stackoverflow.com/questions/9868792/find-out-the-unicode-script-of-a-character), I haven’t evaluated `unicodedata2`, mentioned in that SO post, but maybe it is what we need – and you can check the section of `split.py` below to get acquainted with a method that uses the `regex` package to separate TSVs according to an inputted script property. Perhaps the `regex` package can also be used to ‘discover’ the script property of characters in each TSV?
https://github.com/kylebgorman/wikipron/blob/370d8d1442604c2590def43c502fef5a52c79468/data/src/split.py#L12-L17

kylebgorman:
@agutkin this would be interesting to you, I think.

agutkin:
Out of curiosity - is regex package using ICU under the hood?

kylebgorman:
Just read the code and I don't see any mention of ICU, so I think no.

PR #341, opened 2021-01-30
Simple script detection API.

agutkin:
Implemented simple script detection API based on the unicode script data that is exposed by the [unicodedataplus](https://github.com/iwsfutcmd/unicodedataplus) package ([pypi](https://pypi.org/project/unicodedataplus/)). The implementation computes the histogram of scripts for the given word and returns the dominant script. By default (`strict` mode), all the code-points must belong to the same script.

As proposed in Issue #319, keeping this functionality restricted to the smoke tests only for now.


data/src/split.py
```diff
@@ -1,11 +1,15 @@
 #!/usr/bin/env python
 
+import collections
 import json
 import os
 import sys
+import unicodedataplus
```
kylebgorman, line 7: this is third-party, innit? Should be after `import regex` then.

data/src/split.py
```diff
@@ -17,6 +21,31 @@ def _generalized_check(script: str, word: str) -> bool:
...
+    is enabled, then all the characters must belong to the same script and
+    `(None, None)` is returned on failure.
+
+    Example: "ژۇرنال" -> ("Arabic", 1.0).
+    """
+    script_counts: Dict[str, float] = collections.defaultdict(float)
+    for char in word:
+        script_counts[unicodedataplus.script(char)] += 1.0
+    script_probs = [(s, script_counts[s] / len(word)) for s in script_counts]
+    script_probs = sorted(script_probs, key=lambda x: x[1], reverse=True)
```
kylebgorman, line 39: `lambda x: x[1]` is slow compared to `operator.itemgetter(1)`.

Also, sort in-place with `.sort()` instead of making a copy. (I think it uses quicksort instead of merge-sort then?)

kylebgorman (commented):
This looks good to me but want to hand over to Lucas for a quick look. Haven't thought about the downstream rammifications closely yet.

lfashby (approved):
Wow, this is great! The additions to `split.py` and `test_split.py` are perfect. We’ll have to discuss how to integrate `_detect_best_script_name()`, but this is exactly the sort of functionality I was imagining.

Thanks so much for this.

agutkin:
> Wow, this is great! The additions to `split.py` and `test_split.py` are perfect. We’ll have to discuss how to integrate `_detect_best_script_name()`, but this is exactly the sort of functionality I was imagining.
> 
> Thanks so much for this.

You're very welcome, Lucas! This API is not set in stone, obviously. I suspect we're going to hit difficult cases, which I am not sure are going to be handled correctly - e.g. the use of Vedic religious symbols which I presume are common to all Brahmic scripts, and so on.

kylebgorman:
I'll merge this immediately after #342, assuming that's about ready.

agutkin:
Thanks, guys!

agutkin:
BTW, I presume the names of the scripts defined by the Unicode match (modulo the underscores) the names in ISO 15924. I wonder whether these, in turn, are the names used by Wiktionary.

Issue #315, opened 2021-01-10
[arm] optional material in parentheses


jhdeov:
There are Armenian entries which show optional material in parentheses. For example, any initial sibilant-stop [cluster](https://en.wiktionary.org/wiki/%D5%BD%D5%BF%D5%A1%D5%B6%D5%A1%D5%AC) gets an obligatory schwa in Western Armenian, but an optional one in Eastern Armenian:  [(ə)stɑˈnɑl]. 

The script seems to skip extracting parentheses, so the extracted pronuncation of ստանալ for Eastern is treated as [ə s t a n a l] instead of [(ə) s t a n a l]

kylebgorman:
So "optionality in parentheses", while not an uncommon thing, is not really
mentioned in the Wiktionary pronunciation specs that I see.

I don't know what WikiPron should do here. We try not to do too many
hermeneutics with the pronunciations; if they're not clear, we ought to fix
them upstream and rescrape. I would say this is a good case of "not clear":
it will make sense to linguists of a certain stripe, and perhaps speakers
of the relevant dialects, but it would be much better to just list the
logical possibilities instead of adding a sort of regular language of
pronunciations.

If you are looking at the static scrape in `data/` it's also possible the
Wiktionary data changed since we last ran that, since it seems like
Armenian is under a lot of active development.

jhdeov:
It's understandable that you shouldn't do much hermeneutics. And ideally it should be a thing that the Wiktionary entries themselves should fix. But I don't think it'll be changed any time soon :/ 

For WikiPron, I'm curious if there's a way to get a completely unfilitered extraction from an entry. So for example, when I last ran WikiPron a few days ago, the headword ստանալ [(ə)stanal] was extracted as  [ə s t a n a l]. Is it possible to play with some parameter so that the extracted form is  [(ə) s t a n a l] or  [( ə ) s t a n a l], i.e., the script treats the parentheses as separate "IPA" symbols?

kylebgorman:
Nothing exists for that yet but I suppose it could be built. The extraction
itself isn’t terribly complex. The first version of the script was maybe 30
lines!

jhdeov:
Hi. The readme recommends the use of the flag --no-skip-parens. I used pip (and pip3) right now to get wikipron. But the argument --no-skip-parens is not recognized

```
usage: wikipron [-h] [--phonetic] [--stress] [--no-stress]
                [--syllable-boundaries] [--no-syllable-boundaries]
[...]

PR #343, opened 2021-02-03
Adds a flag for parentheses

m-sean:
* Adds flag to keep parentheses in the transcriptions (enhancement from #315)

### DIFF
--- a/tests/test_wikipron/test_extract.py
+++ b/tests/test_wikipron/test_extract.py
@@ -2,7 +2,7 @@ import pytest
 import requests_html
 
 from wikipron.extract import EXTRACTION_FUNCTIONS
-from wikipron.extract.core import _skip_pron, _handle_parens
+from wikipron.extract.core import _handle_parens, _skip_pron
 from wikipron.extract.default import extract_word_pron_default
 
 
--- a/tests/test_wikipron/test_extract.py
+++ b/tests/test_wikipron/test_extract.py
@@ -2,7 +2,7 @@ import pytest
 import requests_html
 
 from wikipron.extract import EXTRACTION_FUNCTIONS
-from wikipron.extract.core import _skip_pron
+from wikipron.extract.core import _skip_pron, _handle_parens
 from wikipron.extract.default import extract_word_pron_default
 
 
@@ -36,3 +36,14 @@ def test_extraction_functions_have_the_same_signature(func):
 )
 def test__skip_pron(pron, iso639_key, skip_spaces, expected):
     assert _skip_pron(pron, skip_spaces) == expected
+
+
+@pytest.mark.parametrize(
+    "input_pron, skip_parens, expected_pron",
+    [
+        ("mɪskæɹəktəɹ(a)ɪzeɪʃən", True, "mɪskæɹəktəɹaɪzeɪʃən"),
+        ("mɪskæɹəktəɹ(a)ɪzeɪʃən", False, "mɪskæɹəktəɹ(a)ɪzeɪʃən"),
+    ],
+)
+def test__handle_parens(input_pron, skip_parens, expected_pron):
+    assert _handle_parens(input_pron, skip_parens) == expected_pron
--- a/wikipron/cli.py
+++ b/wikipron/cli.py
@@ -104,6 +104,12 @@ def _get_cli_args(args: List[str]) -> argparse.Namespace:
         dest="tone",
         help="removes tones from the transcriptions",
     )
+    parser.add_argument(
+        "--keep-parens",
+        action="store_true",
+        dest="parens",
+        help="include parentheses in the transcriptions",
+    )
     return parser.parse_args(args)
 
 
--- a/wikipron/cli.py
+++ b/wikipron/cli.py
@@ -129,6 +129,5 @@ def _scrape_and_write(config: Config) -> None:
 def main() -> None:
     logging.basicConfig(format="%(levelname)s: %(message)s", level="INFO")
     args = _get_cli_args(sys.argv[1:])
-    print(args)
     config = Config(**args.__dict__)
     _scrape_and_write(config)
--- a/wikipron/cli.py
+++ b/wikipron/cli.py
@@ -105,9 +105,15 @@ def _get_cli_args(args: List[str]) -> argparse.Namespace:
         help="removes tones from the transcriptions",
     )
     parser.add_argument(
-        "--keep-parens",
+        "--skip-parens",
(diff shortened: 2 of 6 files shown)


kylebgorman (commented):
Hey Sean! Good to hear from you.

This loks good to me but I think we may want:

1. a unit test
2. a negative version of the flag? (cf. what @yeonju123 did with `--stress` and `--no-stress`, etc.)

If you have a minute to do either of those that'd be great.

One other mode that one can imagine is to delete the parenthetical content rather than elide the parentheses, but I'm not sure that's useful.

m-sean:
Hi! I’d be happy to add those—should be able to get to them tomorrow or so.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

wikipron/extract/core.py
```diff
@@ -31,8 +31,9 @@ def yield_pron(
         if not m:
             continue
         pron = m.group(1)
-        # Removes parens around various segments.
-        pron = pron.replace("(", "").replace(")", "")
+        # Removes parens around various segments unless --keep-parens is used.
+        if not config.parens:
+            pron = pron.replace("(", "").replace(")", "")
```
jacksonllee, line 36: Agreed that a unit test would be needed. If the key logic update is made here inside `yield_pron`, it might be tricky to write a test against just this. As I saw Kyle's comment wrt possibly adding an option to delete parenthetical material in the future, what do you think about a more future-proof approach along these lines:

* Define a function in `wikipron/extract/core.py` here, something like `_handle_pron(pron: str, config: Config) -> str:`. This function would have the `if not config.parens: pron = pron.replace("(", "").replace(")", "")` logic. Then in `yield_pron`, there would be `pron = _handle_pron(pron, config)` instead.
* If we do want to do other things like dropping parenthetical content later, then `_handle_pron` would be ready to cater for any new logic needed.
* Since `_handle_pron` is a standalone function, it'd be easily testable -- maybe the tests could live in `tests/test_wikipron/test_extract.py`?

wikipron/cli.py
```diff
@@ -104,6 +104,12 @@ def _get_cli_args(args: List[str]) -> argparse.Namespace:
         dest="tone",
         help="removes tones from the transcriptions",
     )
+    parser.add_argument(
+        "--keep-parens",
+        action="store_true",
+        dest="parens",
+        help="include parentheses in the transcriptions",
```
jacksonllee, line 111: I see the other flags are described in the repo's top-level `README.md`, under the "Usage" section. Probably a note on `--keep-parens` would be needed there as well.

wikipron/config.py
```diff
@@ -81,6 +82,7 @@ def __init__(
         )
         self.skip_spaces_word: bool = skip_spaces_word
         self.skip_spaces_pron: bool = skip_spaces_pron
+        self.parens: bool = parens
```
jacksonllee, line 85: What do you think about `self.keep_parens: bool = keep_parens` instead of `self.parens: bool = parens`? Kyle's comment about other potential features related to parens makes me think we probably should be proactive about using more explicit names around earlier.

wikipron/cli.py
```diff
@@ -104,6 +104,12 @@ def _get_cli_args(args: List[str]) -> argparse.Namespace:
         dest="tone",
         help="removes tones from the transcriptions",
     )
+    parser.add_argument(
+        "--keep-parens",
+        action="store_true",
+        dest="parens",
```
jacksonllee, line 110: See my comment re: `keep_parens` vs `parens`

jacksonllee (commented):
Looks good. I left a couple comments.

Please also add an entry to `CHANGELOG.md`.

@kylebgorman As for adding a negative flag, what would that be -- something like `--remove-parens`? What would be its behavior when applied vs not applied, and if interacting with `--keep-parens`?

wikipron/extract/core.py
m-sean: I was having similar thoughts re: the difficulty of testing. This seems like a good solution so I can definitely add it. The only thing is the name—do you think `_handle_paren` might be better, or should we keep it more general in case we want to open it up to other functionality beyond handling parenthetical content?

wikipron/config.py
m-sean: This is reasonable. I just thought `parens` fit in with the existing config attributes, but I can make the change to make it more clear—probably more important.

wikipron/extract/core.py
jacksonllee: > do you think _handle_paren might be better, or should we keep it more general in case we want to open it up to other functionality beyond handling parenthetical content?

The more specific `_handle_parens` is a great idea. If we need other "handle X" functions in the future, we can always add and test them separately as needed.

kylebgorman: +1 to adding a function for this.

wikipron/cli.py
kylebgorman: +1

wikipron/config.py
kylebgorman: In fact maybe `skip_parens` sould be better yet?

m-sean: Good point. I'll reverse the logic and rename the flags `--skip-parens` and `--no-skip-parens` to keep it consistent with the other flags.

jacksonllee (approved):
LGTM!

tests/test_wikipron/test_extract.py
```diff
@@ -2,7 +2,7 @@
 import requests_html
 
 from wikipron.extract import EXTRACTION_FUNCTIONS
-from wikipron.extract.core import _skip_pron
+from wikipron.extract.core import _skip_pron, _handle_parens
```
kylebgorman, line 5: Extremely minor but could you sort this lexicographically? thus `_handle_parens` first?

kylebgorman (commented):
Thanks Sean. One comment, plus could you add this to the top-level README as per Jackson's request? Then this is ready to go in.

tests/test_wikipron/test_extract.py
m-sean: sure! Also there's an addition to the README at line 78 you must've missed. Let me know if you'd like me to make any adjustments before merging

kylebgorman (approved):
LGTM

tests/test_wikipron/test_extract.py
kylebgorman: Sorry I missed that. Thanks!

PR #344, opened 2021-02-04
Create generate_phone_summary.py

yeonju123:
This script automatically generates phone_summary, which has similar structure as language_summary. This script is written  based on `generate_summary.py` .

The output of `generate_phone_summary.py` is a TSV file instead of README.md, since README.md is already used. I will change the output path after we discuss what would be the good path for the output.


kylebgorman (commented):
Thanks for this Yeonju. A few minro things:

- I don't like the naming conventions `generate_summary.py` and `generate_phones_summary.py`. In general I hate that one is more specific than the other. Can we do something different? Maybe rename `generate_summary.py` to `generate_tsv_summary.py` and update the relevant documentation?
- You should check in the generated README with this PR. I'd love to preview it...
- Please update the changelist as requested in the PR description, commit that, and click the checkbox in the PR description.
- The `black` test is failing. Reformat the files with `black -l79 *.py` and commit any changed files.

I'm going to add another set of eyeballs to this too.

yeonju123:
Thank you for the comments! I will work on that this evening :)

lfashby (commented):
Hey Yeonju, 
I have a few comments related to only outputting a TSV. I guess we should decide whether we want this to output a TSV or a README.md or both, but if we just want to output only a TSV then see my comments below.  If we decide to output both a README.md and a TSV then you can disregard most of my comments.

data/src/generate_phone_summary.py
```diff
@@ -0,0 +1,96 @@
...
+                values = values.replace(" |", ",")
+            name += f" ({values})"
+    return name
+
+
+def main() -> None:
+    with open(LANGUAGES_PATH, "r", encoding="utf-8") as source:
+        languages = json.load(source)
+    readme_list = []
+    languages_summary_list = []
```
lfashby, line 44: We can get rid of this.

data/src/generate_phone_summary.py
```diff
@@ -0,0 +1,96 @@
...
+    path = "../phones"
+    modifiers = ["dialect", "script"]
```
lfashby, line 46: This can just be `modifiers = ["dialect"]` since we are only concerned with transcriptions here.

data/src/generate_phone_summary.py
```diff
@@ -0,0 +1,96 @@
...
+
+    for file_path in os.listdir(path):
+        # Filters out README.md.
+        if file_path.endswith(".md"):
```
lfashby, line 50: The `phone_summary.tsv` that this outputs is currently going into the `phones` directory, so when you run this code a second time (when `phone_summary.tsv` is already in `phones`) we'll want to skip it as well otherwise it'll throw off the substring stuff below.

Maybe:
```
if file_path.endswith(".md") or file_path.endswith(".tsv"):
```

data/src/generate_phone_summary.py
```diff
@@ -0,0 +1,96 @@
...
+            languages[iso639_code], file_path, modifiers
+        )
+        row = [
+            iso639_code,
+            languages[iso639_code]["iso639_name"],
+            wiki_name,
+            transcription_level,
+            num_of_entries,
+        ]
+        readme_list.append([f"[phone](phones/{file_path})"] + row)
```
lfashby, line 68: If we're only outputting a TSV this can just be: `readme_list.append([file_path] + row)`

data/src/generate_phone_summary.py
```diff
@@ -0,0 +1,96 @@
...
+    # Sorts by Wiktionary language name, with phonemic entries before phonetic
+    # ones.
+    readme_list.sort(key=_wiki_name_and_transcription_level)
+    # Writes the README.
+    with open(PHONES_SUMMARY_PATH, "w", encoding="utf-8") as sink:
```
lfashby, line 73: If we're only outputting a TSV this should be:
```
    with open(PHONES_SUMMARY_PATH, "w", encoding="utf-8") as sink:
        tsv_writer_object = csv.writer(
            sink, delimiter="\t", lineterminator="\n"
        )
        tsv_writer_object.writerows(readme_list)
```
And everything within the `with open(PHONES_SUMMARY_PATH, "w", encoding="utf-8") as sink:` block can be removed.

yeonju123:
Thank you for the detailed comments. These are really helpful :)
I am still not sure of whether we want to have both README.md and TSV or
only one of them. Initially Kyle suggested to generate a README.md, but we
already have one. We may be just able to add the summary table on the top
or below the current README.md, but I am not sure of an efficient way to do
it.
What output type do you think would be good?

kylebgorman:
One point of clarification. Obviously you can only have one file of any given name in any given directory. However, you can have a `README.md` in every directory in a GitHub project and GitHub will render it correctly. The file `data/phones` currently has a manually-generated `README.md` but that could just be named something else like "HOWTO.md" or something; or it could be prepended to the autogenerated `README.md` file that this script generates.

One reason I asked Lucas to generate both a TSV file and a `README.md` is that they're for different purposes. I used the TSV file to build tables for the Lee et al. 2020 paper, and `README.md` is for users to read. I'd say we probably want both here too. Then I propose:

* phone summary script: `generate_phones_summary.py`
* TSV file it generates: `phones_summary.tsv`
* README file it generates: `../phones/README.md`
* TSV summary script: `generate_tsv_summary.py`
* TSV file it generates: `tsv_summary.py`
* README file it generates: `../tsv/README.md`

lfashby:
I agree with Kyle's proposal, although I think the README and TSV created by `generate_summary.py` should remain in `data`. I don't think the (`generate_summary.py`) README should be placed in `data/tsv`, as that would mean (on Github) having to scroll to the bottom of a huge list of files to look at the README.

kylebgorman:
Oops sorry, I agree with Lucas, I forgot about that problem.

yeonju123:
Thank you for clarifying. I also agree with both of you. Generating a table
may be useful at some point. Then, I will fix the code and push it again,
based on Lucas's suggestion. So far the list of phones is short, so
prepending it to the current README.md would be fine. But we expect the
list will be bigger, it may be better to have separate *md files. For now,
I will just creat two *.mds (HOWTO, README). If you have different opinions
about this, let me know.

ajmalanoski:
To add my two cents, I'm partial to having the how-to and the phones summary both in the README (with the how-to on top), just to lower the chance of someone missing the directions.

ajmalanoski (commented):
I don't have any additions to the comments that have already been made. Other than those, LGTM.

yeonju123:
@ajmalanoski That's a good point. Um.. but we have the file listed in phones. If we have both HOWTO and summary together in README.md, wouldn't it require too much scrolling-down? Um.. currently, the script is generating a separate README.md but tomorrow, I will try to put HOWTO and README together and check whether it is too long :)

yeonju123:
Oh.. the tests are failed because language_summary.tsv does not exist anymore...

lfashby:
I should have thought of this but to get the tests to pass you'll need to change:
https://github.com/kylebgorman/wikipron/blob/413c2b34a019ff897f4dc6db6b1212d85fe198e6/tests/test_data/test_summary.py#L6
to `_SUMMARY = os.path.join(_REPO_DIR, "data/tsv_summary.tsv")`. There are also some comments in that file (`test_summary.py`) mentioning 'languages_summary.tsv' and 'generate_summary.py' which you could change to their new names if you fancy.

It'd be great to have the instructions and the phones summary together, but I think it's just much easier to have them be separate. If they were together in one markdown file we'd have to implement a way of overwriting the phones summary section with the output of `generate_phone_summary.py` while leaving the instructions section unchanged. Perhaps there is an easy way of doing that but I can't think of it currently.

kylebgorman:
> It'd be great to have the instructions and the phones summary together, but I think it's just much easier to have them be separate. If they were together in one markdown file we'd have to implement a way of overwriting the phones summary section with the output of `generate_phone_summary.py` while leaving the instructions section unchanged. Perhaps there is an easy way of doing that but I can't think of it currently.

You can just link from one to the other, if you want to. It'll look something like:

`See the [HOWTO](HOWTO.md) for more information.`

yeonju123:
This PR has conflicts because `languages_summary.tsv` is not present anymore. Should I just follow the recommendation from github to resolve this?

kylebgorman:
Tests now passing, thanks for this.

kylebgorman (approved):
LGTM.

PR #356, opened 2021-02-10
[slv] Fixes Slovenian normalization. Re-scrapes Slovenian.

ajmalanoski:

Also adds a script to change a file's unicode normalization


data/src/normalize.py
```diff
@@ -0,0 +1,39 @@
+import argparse
+
+from tempfile import TemporaryFile
```
kylebgorman, line 3: Please use absolute imports like the rest of the library.

ajmalanoski:
Oh, I just remembered I forgot to do the postprocessing and summaries. I'll fix that right now

data/src/normalize.py
```diff
@@ -0,0 +1,39 @@
...
+from unicodedata import normalize
+
+'''
```
kylebgorman, line 6: 1. This has to go above any imports for Python to understand that it's documentatino.
2. Don't include the program name: that's self-evident.
3. The first sentence should be a brief <80 explanation, then you go into normal prose. E.g.:

```
"""In-place Unicode normalization.

This script takes a file and normalizes it in-place...
"""
```

4. Use " not '. " is a better string-quote character.

data/src/normalize.py
```diff
@@ -0,0 +1,39 @@
...
+normalize.py
+
+Takes a file and normalizes it "in place." In order to avoid the issues of
+reading and writing to the same file at the same time, this script puts the 
+normalized version of the file argument in a tempfile, then uses that tempfile
+to rewrite the original file.
+'''
+
+def main(args: argparse.Namespace) -> None:
```
kylebgorman, line 15: There is way too much whitespace in this little function. Use a four-space indent (as is the fashion) and get rid of the blank lines. I think blank lines in a single function make the function boundary very hard to discern. If you're tempted to paragraph-ify things, put a comment in instead, or write shorter functions.

data/src/normalize.py
```diff
@@ -0,0 +1,39 @@
...
+		for line in rf:
+			tf.write(normalize(args.norm, line))
+
+	tf.seek(0)
+
+	with open(args.file, 'w') as wf:
+		for line in tf:
+			wf.write(line)
+
+	tf.close()
```
kylebgorman, line 28: I think you can once again use a `with`-block for `tf` instead of `.close`.

data/src/normalize.py
```diff
@@ -0,0 +1,39 @@
...
+normalized version of the file argument in a tempfile, then uses that tempfile
+to rewrite the original file.
+'''
+
+def main(args: argparse.Namespace) -> None:
+	tf = TemporaryFile(mode='w+')
+
+	with open(args.file, 'r') as rf:
+		for line in rf:
+			tf.write(normalize(args.norm, line))
```
kylebgorman, line 20: Prefer `print` to `write`. (it has an option to not add newlines if that's what you want.)

data/src/normalize.py
```diff
@@ -0,0 +1,39 @@
...
+		for line in tf:
+			wf.write(line)
+
+	tf.close()
+
+
+
+if __name__ == "__main__":
+	parser = argparse.ArgumentParser(description="Changes a file's unicode normalization")
+	# File to normalize
```
kylebgorman, line 34: These two comments are redundant with the documentation.

data/src/normalize.py
```diff
@@ -0,0 +1,39 @@
...
+	parser.add_argument("file", help="The file to modify")
+	# Normalization type
+	parser.add_argument("norm", choices=["NFC","NFD","NFKC","NFKD"], help="The type of normalization desired")
```
kylebgorman, line 37: Need space after `.`, run `black` and `flake` on these? I would also use the term "normalization form".

data/src/normalize.py
```diff
@@ -0,0 +1,39 @@
...
+	with open(args.file, 'w') as wf:
+		for line in tf:
+			wf.write(line)
+
+	tf.close()
+
+
+
+if __name__ == "__main__":
+	parser = argparse.ArgumentParser(description="Changes a file's unicode normalization")
```
kylebgorman, line 33: If you put the docstring in the format I recommended you can do:

`description=__doc__`

And it will automatically inherit that short one-line description.

data/src/normalize.py
```diff
@@ -0,0 +1,39 @@
...
+	# File to normalize
+	parser.add_argument("file", help="The file to modify")
```
kylebgorman, line 35: I realize this isn't very intuitive but the autogenerated help lines start with a lowercase, so the ones you type should also do so. Thus `help="file to modify"`, etc. I also recommend not using `file`. It's a built-in in Python.

data/src/normalize.py
```diff
@@ -0,0 +1,39 @@
...
+	# Normalization type
+	parser.add_argument("norm", choices=["NFC","NFD","NFKC","NFKD"], help="The type of normalization desired")
+
```
kylebgorman, line 38: Would not put a space here. But do put a newline at the end...that's what that red `(-)` that GitHub puts at the end of the thing is complaining about.

data/src/normalize.py
```diff
@@ -0,0 +1,39 @@
...
+	tf.seek(0)
+
+	with open(args.file, 'w') as wf:
+		for line in tf:
+			wf.write(line)
+
+	tf.close()
+
+
```
kylebgorman, line 30: 2 blank lines between functions, never more or less.

kylebgorman (commented):
Sorry, but lots of finicky complaints about `src/normalize.py`; I'm trying to preserve a standard style (which is roughly PEP-8) throughout.

data/src/normalize.py
ajmalanoski: Is there a practical reason for this, or is it just a stylistic thing? Just wondering.

kylebgorman: I suppose it's stylistic at this point: in the 16-odd years I've been doing this, I've almost never seen `write` used.

data/src/normalize.py
ajmalanoski: Is this just stylistic, or is there a practical reason for including a newline at the end?

kylebgorman: I am not sure what to call it, but GitHub kinda complains if you don't so I do.

kylebgorman:
Can you run `black -l79` over the `normalize.py` and try again? That seems to be the broken test.

ajmalanoski:
I don't know why the black test is still failing. Every line is under 79 characters now.

kylebgorman:
Strange. It checks (and can change) things other than line length, so maybe something else is at play.

ajmalanoski:
Ah, I hadn't seen your previous comment when I wrote my last comment. I ran the `black` command and it's working now.

kylebgorman (approved):
LGTM. I'll let you push the big green button...it's very satisfying.

PR #358, opened 2021-02-12
Update `data/src/normalize.py` to be more efficient

ajmalanoski:

Following @kylebgorman 's suggestion, I modified the file to be somewhat more efficient. Instead of looping over the tempfile to overwrite the input file, the file now "moves" the contents of the tempfile over the input file, which I am told is an _O(1)_ operation.

I want to record for posterity one thing that was unintuitive to me. On line 19, when the NamedTemporaryFile is created, I had to set `delete=False`. With `delete=True` (the default), Python raised a `FileNotFoundError` at line 22 `shutil.move(tf.name, args.file_name)`. This problem was resolved by setting `delete=False`.


ajmalanoski:
I just noticed that I put the wrong path to `data/src/normalize.py` when I updated the changelog for #356, so I fixed that right now.

kylebgorman (approved):
LGTM.

Issue #360, opened 2021-02-18
wikipron/data/src reorganization


lfashby:
[`wikipron/data/src`](https://github.com/kylebgorman/wikipron/tree/master/data/src) is host to a lot of different files that do a lot of different things in different places. The unorganized state of `wikipron/data/src` may be off-putting to new/prospective contributors who don’t want to wade into this large collection of files to try and get their bearings.

I have a few suggestions for things we can do to reorganize `wikipron/data/src`, though I’d also like suggestions for how (or even if) `wikipron/data/src` should be reorganized from those who contributed some of these files.

A few things come to mind immediately:
- We wrote `postprocess` but for whatever reason only call `split.py` from within it. We should call `generate_tsv_summary.py` from `postprocess` as well and maybe put both `split.py` and `generate_tsv_summary.py` in an `src/postprocess_helpers` subdirectory.
- All of our `.json` files should go in a subdirectory (or at least all of them that are not `languages.json`).
- The “Big Scrape” Scripts README is out of date, we no longer retry languages 10 times.
- The scripts which have to do with the phones stuff should be in there own subdirectory? There are quite a few of them now.
- `covering_grammar.py` seems a bit homeless. I don’t think we’ve used it.

I’m not sure if putting things in subdirectories of `src` is a great solution, so any alternatives would be welcome.

kylebgorman:
> [`wikipron/data/src`](https://github.com/kylebgorman/wikipron/tree/master/data/src) is host to a lot of different files that do a lot of different things in different places. The unorganized state of `wikipron/data/src` may be off-putting to new/prospective contributors who don’t want to wade into this large collection of files to try and get their bearings.
>
> I have a few suggestions for things we can do to reorganize `wikipron/data/src`, though I’d also like suggestions for how (or even if) `wikipron/data/src` should be reorganized from those who contributed some of these files.

I agree this needs to be reorganized. It's a mess.

> A few things come to mind immediately:
> 
> * We wrote `postprocess` but for whatever reason only call `split.py` from within it. We should call `generate_tsv_summary.py` from `postprocess` as well and maybe put both `split.py` and `generate_tsv_summary.py` in an `src/postprocess_helpers` subdirectory.

+1.

The things people need to call: `scrape.py` and `postprocess` should live in some top-level directory, put everything else in a subdirectory maybe.

> * All of our `.json` files should go in a subdirectory (or at least all of them that are not `languages.json`).

[...]

PR #361, opened 2021-02-18
Updated README to remove reference to retries.

kylebgorman:
Part of #360.

Also reflowed the entire file using:

    pandoc -f markdown -t markdown --columns=80

### DIFF
--- a/data/src/README.md
+++ b/data/src/README.md
@@ -2,23 +2,22 @@
 ====================
 
 [`scrape.py`](scrape.py) calls WikiPron's scraping functions on all Wiktionary
-languages with over 100 entries. If a `.phones` file is present for a given language,
-the process will generate an additional filtered file only containing the permitted
-phones/phonemes.
+languages with over 100 entries. If a `.phones` file is present for a given
+language, the process will generate an additional filtered file only containing
+the permitted phones/phonemes.
 [`generate_tsv_summary.py`](generate_tsv_summary.py) generates a
-[README](../README.md) and a [TSV](../tsv_summary.tsv) with selected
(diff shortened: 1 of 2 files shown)


data/src/README.md
```diff
@@ -48,9 +47,9 @@ Steps used to update the dataset
...
 --------------------------------------------------
 
-By default, if [`scrape.py`](scrape.py) cannot successfully complete a scrape of
-an entire language in 10 retries, it will log the language, remove the
-incomplete data scraped from that language, and move on to the next language in
+By default, if [`scrape.py`](scrape.py) cannot complete the scrape of a
+language, it will log the language, remove the incomplete data scraped from that
+language, and move on to the next language in
 [`languages.json`](languages.json). There may therefore be a few languages that
 you need to run again when the big scrape finishes. These are the steps to
```
lfashby, line 54: When I added the restart functionality I removed the logging of incomplete scrapes and the deletion of incomplete data from `src/scrape.py`. Other than a user ctrl + c-ing or the user's internet dropping (or Wiktionary itself dropping off the internet somehow) there isn't any reason `src/scrape.py` shouldn't complete a scrape.

I think this section can just begin with: "These are the steps to follow should you need to run the big scrape scripts for a smaller set of languages:"

lfashby (commented):
Just a small comment about the new behavior of `src/scrape.py` now that we have the restart functionality in `wikipron/scrape.py`.

data/src/README.md
kylebgorman: Makes sense, done.

lfashby (approved):
LGTM, thanks for this.

PR #364, opened 2021-02-23
Automatically update languages.json 

### LINKED ISSUE #189: [got] Handling of reconstructed (unattested) spellings
Historic languages like Gothic (`got`) contain some pronunciations derived from the [reconstructed](https://github.com/kylebgorman/wikipron/blob/master/data/wikipron/tsv/got_phonemic.tsv) spellings. This reside in `reconstruction:` [namespace](https://en.wiktionary.org/wiki/Wiktionary:Reconstructed_terms)  which ends up in the corresponding spelling prefixes, e.g. `reconstruction:gothic/𐌷𐌻𐌰𐌸𐌰𐌽`.

Some strategy needs to be worked out on how to handle such cases, perhaps by introducing a flag to weed out or keep the reconstructed entries.

### LINKED ISSUE #319: Automatic script inspection

cgibson6279:
Closes #189. Closes #319.


data/src/languages_update.py
```diff
@@ -0,0 +1,127 @@
+#!/usr/bin/env python
+
+import collections
+import json
+import operator
+import os
+import sys
+
+import unicodedataplus
+from data.src.codes import LANGUAGES_PATH, TSV_DIRECTORY_PATH
```
kylebgorman, line 10: newline between here. `Unicodedata is third-party`, `data` is part of this package.

data/src/languages_update.py
```diff
@@ -0,0 +1,127 @@
...
+
+from typing import Dict, Tuple, Union
```
kylebgorman, line 12: `typing` imports go after standard module imports

data/src/languages_update.py
```diff
@@ -0,0 +1,127 @@
...
+) -> Union[Tuple[str, float], None]:
+    """Returns the most likely script name (rather than ISO 15924 code) the
+    word belongs to along with the corresponding confidence expressed as a
+    maximum likelihood estimate computed over the `word` sample. If `strict`
+    is enabled, then all the characters must belong to the same script and
+    `(None, None)` is returned on failure.
+
+    Example: "ژۇرنال" -> ("Arabic", 1.0).
+    """
+    script_counts: Dict[str, float] = collections.defaultdict(float)
```
kylebgorman, line 25: Use this more specific type instead: https://docs.python.org/3/library/typing.html#typing.DefaultDict

data/src/languages_update.py
```diff
@@ -0,0 +1,127 @@
...
+    for char in word:
+        script_counts[unicodedataplus.script(char)] += 1.0
```
kylebgorman, line 27: I'm slightly confused why counts are fractional (1.0) instead of integral.

data/src/languages_update.py
```diff
@@ -0,0 +1,127 @@
...
+        #   https://www.unicode.org/Public/13.0.0/ucd/Scripts.txt
+        return script_probs[0][0], script_probs[0][1]
+
+
+def _remove_duplicates(script_dict: Dict[str, str]) -> Dict[str, str]:
+    '''
+    If a values in lang["script"] appears more than once, the [key:value] pair that does not conform to ISO unicode
+    entries returned from unicodedataplus.property_value_aliases['script'] is deleted.
+
+    eg)
```
kylebgorman, line 44: what is `eg)`?

This is kind of a long example, I wonder if it would be better as a test.

data/src/languages_update.py
```diff
@@ -0,0 +1,127 @@
...
+        "iso639_name": "Azerbaijani",
+        "wiktionary_name": "Azerbaijani",
+        "wiktionary_code": "az",
+        "casefold": true,
+        "script": {
+            "latn": "Latin",
+            "cyrl": "Cyrillic",
+            "arab": "Arabic"
+        }
+
```
kylebgorman, line 70: Cut the trailing whitespace here

data/src/languages_update.py
```diff
@@ -0,0 +1,127 @@
...
+
+    '''
+    remove = []
+
+    for key, value in script_dict["script"].items():
+        value = value.replace(" ", "_")
+        if not ''.join(unicodedataplus.property_value_aliases['script'][value]).lower() == key:
```
kylebgorman, line 77: instead of `if not x == y:` prefer the more to-the-point "if x != y:`.

data/src/languages_update.py
```diff
@@ -0,0 +1,127 @@
+#!/usr/bin/env python
```
kylebgorman, line 1: Please give this a module-level docstring.

data/src/languages_update.py
```diff
@@ -0,0 +1,127 @@
...
+        return None
+    else:
+        # The script names in Unicode data tables have underscores instead of
+        # whitespace to enable parsing. See:
+        #   https://www.unicode.org/Public/13.0.0/ucd/Scripts.txt
+        return script_probs[0][0], script_probs[0][1]
+
+
+def _remove_duplicates(script_dict: Dict[str, str]) -> Dict[str, str]:
+    '''
```
kylebgorman, line 40: Please use `"""` rather than `'''` throughout, and also `"`" rather than `'` except if there's some good reason to prefer the other.

data/src/languages_update.py
```diff
@@ -0,0 +1,127 @@
...
+        for file in os.listdir(tsv_path):
+            if file.endswith('.tsv'):
+                iso639_code = file[:file.index("_")]
+                lang = languages[iso639_code]
+                with open(f'{tsv_path}/{file}', "r", encoding="utf-8") as f:
+                    for line in f:
+                        try:
+                            word = line.split("\t", 1)[0]
+                            script, prob = _detect_best_script_name(word)
+                            if not "script" in lang:
```
kylebgorman, line 101: prefer `if x not in y:` to `if not x in y:`. Why? It's more English-like, if nothing else.

data/src/languages_update.py
```diff
@@ -0,0 +1,127 @@
...
+                                lang["script"] = {}
+                            # use property_value_aliases to get ISO 15924 code
```
kylebgorman, line 103: `# Uses property_value_aliases to get ISO-15924 codes.`

Ideally these should look like real sentences.

data/src/languages_update.py
```diff
@@ -0,0 +1,127 @@
...
+                            if not script in lang["script"]:
+                                lang["script"][''.join(
+                                    unicodedataplus.property_value_aliases['script'][script]).lower()] = script.replace(
+                                    "_", " ")
+                            _remove_duplicates(lang)
+                        except TypeError as error:
```
kylebgorman, line 109: Reading this I'm not sure why a `TypeError` would be raised, and it makes me nervous. Maybe worth a comment.

data/src/languages_update.py
```diff
@@ -0,0 +1,127 @@
...
+                            pass
+        json_object = json.dumps(languages, indent=4)
+
+        with open(output_path, "w", encoding="utf-8") as lang_source:
+            lang_source.write(json_object)
+
+
+def main():
+    '''
```
kylebgorman, line 118: I wouldn't put a docstring on a main, but FYI the `"""`s and the one-line comment should be on the same line or it's not a one-line comment anymore.

data/src/languages_update.py
```diff
@@ -0,0 +1,127 @@
+#!/usr/bin/env python
+
```
kylebgorman, line 2: Coudl you please remove all in-function whitespace? It makes it hard to parse when the function ends. If you really want to chunk it somehow, add a comment instead.

data/src/split.py
```diff
@@ -44,7 +44,84 @@ def _detect_best_script_name(
         # The script names in Unicode data tables have underscores instead of
         # whitespace to enable parsing. See:
         #   https://www.unicode.org/Public/13.0.0/ucd/Scripts.txt
-        return script_probs[0][0].replace("_", " "), script_probs[0][1]
+        return script_probs[0][0], script_probs[0][1]
+
+
+def _remove_duplicates(script_dict: Dict[str, str]) -> Dict[str, str]:
```
kylebgorman, line 50: This function has been repeated from `languages_update.py`. Is that intended?

data/src/split.py
```diff
@@ -44,7 +44,84 @@ def _detect_best_script_name(
...
+    for key, value in script_dict["script"].items():
+        value = value.replace(" ", "_")
+        if not ''.join(unicodedataplus.property_value_aliases['script'][value]).lower() == key:
+            remove.append(key)
+    for i in remove:
+        del script_dict["script"][i]
+    return script_dict
+
+
+def _update_languages_json(tsv_path: str, LANGUAGES_PATH: str) -> None:
```
kylebgorman, line 95: Repeated function: stay DRY.

data/src/split.py
```diff
@@ -60,10 +137,16 @@ def _iterate_through_file(
 
 def main() -> None:
     tsv_path = sys.argv[1]
+
+    #updates languages.json before splitting files
```
kylebgorman, line 141: `# Updates languages.json before splitting files.`

kylebgorman (commented):
Thanks for this Cameron, this is conceptually exactly what I we needed here.

1. I have some comments about style below. About half are ones you'd get from anybody following PEP-8, the rest are either my specific neuroses or "package style".
2. I am also concerned how so much code in `languages_update.py` is repeated in `split.py`. Can you put the relevant pieces in a separate module and import them instead?
3. I think you need to add `unicodedataplus` to the requirements.txt file in `data/`. Initially, I'd suggest not specifying a version; if we find we need a specific version (or a version range, or a version inequality...) we can add that later.
4. We should discuss whether we want to add the language codes to the name of "big scrape" TSV file. @agutkin you've been thinking about this...any thoughts?
5. If we do that we should probably enrich the autogenerated tables to include script information.

kylebgorman:
You'll also have to run `black -79` over everything and make `mypy` tests pass locally.

kylebgorman:
Please add "closes #319" to the description of the PR. GitHub will then automatically close that issue and link it against this PR when the PR is merged.

lfashby:
Hey Cameron, 
Thanks for this. I won't really have time to get to this until tomorrow night/Thursday, but I'll try and leave some comments if I'm free before then.

data/src/languages.json
```diff
@@ -359,85 +496,130 @@
...
+        }
     },
     "ger": {
         "iso639_name": "German",
         "wiktionary_name": "German",
         "wiktionary_code": "de",
-        "casefold": true
+        "casefold": true,
+        "script": {
+            "zyyy": "Common",
```
agutkin, line 593: We should double-check which words trigger the "Common" script - it's likely that we need to fold in "Common" script into "Latin" and "Cyrillic" cases.

data/src/languages.json
```diff
@@ -449,103 +631,157 @@
...
     },
     "haw": {
         "iso639_name": "Hawaiian",
         "wiktionary_name": "Hawaiian",
         "wiktionary_code": "haw",
-        "casefold": true
+        "casefold": true,
+        "script": {
+            "latn": "Latin",
+            "zyyy": "Common"
```
agutkin, line 655: Same as my previous comment: May need to be folded at the implementation level...

data/src/languages.json
```diff
@@ -585,7 +827,12 @@
         "iso639_name": "Kashmiri",
         "wiktionary_name": "Kashmiri",
         "wiktionary_code": "ks",
-        "casefold": false
+        "casefold": false,
+        "script": {
+            "arab": "Arabic",
+            "zinhqaai": "Inherited",
```
agutkin, line 833: Interesting. How's this triggered?

data/src/languages.json
```diff
@@ -665,97 +936,147 @@
...
-        "casefold": true
+        "casefold": true,
+        "script": {
+            "latn": "Latin"
+        }
     },
     "khb": {
-        "iso639_name": "Lü",
-        "wiktionary_name": "Lü",
+        "iso639_name": "L\u00fc",
```
agutkin, line 1054: Wrong output mode or wrong string type in format statement?

data/src/languages.json
```diff
@@ -1385,24 +1964,33 @@
         "skip_spaces_word": false,
         "skip_spaces_pron": false,
         "dialect": {
-            "hanoi": "Hà Nội",
-            "hcmc": "Hồ Chí Minh City",
-            "hue": "Huế",
-            "tc": "Vinh, Thanh Chương",
-            "ht": "Hà Tĩnh"
+            "hanoi": "H\u00e0 N\u1ed9i",
```
agutkin, line 1967: Strings not output correctly.

data/src/languages_update.py
```diff
@@ -0,0 +1,127 @@
...
+    if strict and len(script_probs) != 1:
+        return None
+    else:
+        # The script names in Unicode data tables have underscores instead of
+        # whitespace to enable parsing. See:
+        #   https://www.unicode.org/Public/13.0.0/ucd/Scripts.txt
+        return script_probs[0][0], script_probs[0][1]
+
+
+def _remove_duplicates(script_dict: Dict[str, str]) -> Dict[str, str]:
```
ajmalanoski, line 39: Is `Dict[str, str]` the right type for `script_dict`? At line 75, it looks like `script_dict["script"]` is itself a string-to-string dictionary, not a string.

data/src/languages.json
cgibson6279: yeah, I forgot to include `ensure_ascii=False` in json.dumps, but I ran it with that param, and that fixed it.

data/tsv/egy_latn_phonemic.tsv
```diff
@@ -2051,12 +1995,6 @@ tr	t ɛ r
 ttj	t a t i j
 ttj	t ɛ t i
 tw	t uː
-tw.j	t uː ʔ i
```
agutkin, line 2054: I suspect it's the symbols like "." that trigger the "Common" script - we'll need to modify the logic and extend the testcase to not filter those.

data/src/languages_update.py
ajmalanoski, line 109: Since the variable `error` is not actually used in the `except` clause, I believe you can just put `except TypeError:`

data/src/split.py
```diff
@@ -44,7 +44,84 @@ def _detect_best_script_name(
...
+                    for line in f:
+                        try:
+                            word = line.split("\t", 1)[0]
+                            script, prob = _detect_best_script_name(word)
+                            if not "script" in lang:
+                                lang["script"] = {}
+                            # use property_value_aliases to get ISO 15924 code
+                            if not script in lang["script"]:
+                                lang["script"][''.join(
+                                    unicodedataplus.property_value_aliases['script'][script]).lower()] = script.replace(
```
ajmalanoski, line 117: It looks like this long piece of code — `''.join(unicodedataplus.property_value_aliases['script'][<value>]).lower()` — occurs several times throughout the code. I think it improve readability if you made a helper function for this, like 
```
def _get_alias(value: str) -> str:
    return ''.join(unicodedataplus.property_value_aliases['script'][value]).lower()
```
It would also be an opportunity to explain what this is doing, because I don't know what these "property value aliases" are.

data/tsv/egy_latn_phonemic.tsv
lfashby: I think that's correct. One thing to be aware of is we will get entries like the following in a scrape of English:
```
++	p l ʌ s p l ʌ s
@	i t ʃ æ t
```
Or in a scrape of Chinese:
```
🌶	l ä
🍋	s w ä n ⁵ ⁵
```
(All classed as Common)
I suppose we have to decide whether we want such entries in our scraped data, if we decide we don't then, at least for English, I think we'll need a more complex solution than folding Common into Latin.

data/src/split.py
```diff
@@ -44,7 +44,84 @@ def _detect_best_script_name(
...
+
+def _update_languages_json(tsv_path: str, LANGUAGES_PATH: str) -> None:
+    '''
+    Detects and identifies all unicode scripts present in a tsv file
+    and updates languages.json to reflect updated ["script"]
+    entries for each language in languages.json
+    '''
+    with open(LANGUAGES_PATH, "r", encoding="utf-8") as lang_source:
+        languages = json.load(lang_source)
+        for file in os.listdir(tsv_path):
```
lfashby, line 103: I fear this might make `postprocess` (where `split.py` is called) take an unreasonably long time.

`postprocess` itself is already running through all the files in the `tsv` directory and calling `split.py` on each. Given the line above, `postprocess` is going to be running through every file in `tsv` and for each file, `split.py` is going to run through every file in `tsv`. Scary stuff.

I think we should either update `languages.json` before `postprocess` runs through `tsv` (meaning pull this functionality out of `split.py` and put it into a separate file which is run at the top of `postprocess`), or just update this function to rewrite `languages.json` one language at a time as each language is passed to `split.py`.

data/src/split.py
lfashby: I think someone mentioned this but `_remove_duplicates()` would function better as a test in `test_data` that enforces our ISO 15924 script key convention in `languages.json`.

data/src/split.py
cgibson6279: I think updating the languages.json before postprocess makes the most sense here. `languages_update.py` runs fairly quickly on its own (no longer than a minute or two or my machine, but I can log it), and it seems more "tidy" approach to me, but what do I know? lol.

data/src/split.py
cgibson6279: makes sense to me. will do.

data/tsv/egy_latn_phonemic.tsv
cgibson6279: because `_detect_best_script_name()` returns a probability as well, I'd imagine that `++` would return 1.0 probability for 'Common' on this symbol and a value less than 1.0 for something like 'tw.j'. I didn't use that probability for anything in the script, but we might be able to use it to help inform classification? I'll play around with things and see if that's even true to begin with.

cgibson6279:
These issues seems fairly straight forward to resolve for the most part, but before I get to work on things, what's the consensus on whether the `languages.json` should be rewritten by a separate script in `postprocess` executed before `split.py ` or as a function of `split.py` while it's running?

data/src/languages_update.py
```diff
@@ -1,15 +1,19 @@
...
+"""
 import collections
 import json
 import operator
 import os
-import sys
-
 import unicodedataplus
+
+from typing import Tuple, Union, DefaultDict, Dict
```
agutkin, line 13: Can you sort the imports? i.e.

   from typing import DefaultDict, Dict, Tuple, Union

data/src/languages_update.py
```diff
@@ -35,92 +39,63 @@ def _detect_best_script_name(
...
-            "arab": "Arabic"
-        }
-
+    Example: "Arabic" -> "arab"
+    """
+    return ''.join(unicodedataplus.property_value_aliases['script'][value]).lower()
 
-    '''
+def _remove_mismatch_ids(script_dict: Dict[str, str]) -> Dict[str, str]:
+    """If a values in lang["script"] appears more than once, the [key:value] pair that does not conform to ISO unicode
```
agutkin, line 51: Can you make the documentation respect the 80 characters in line limit please?

data/src/languages_update.py
```diff
@@ -35,92 +39,63 @@ def _detect_best_script_name(
...
                             word = line.split("\t", 1)[0]
                             script, prob = _detect_best_script_name(word)
-                            if not "script" in lang:
+                            if "script" not in lang:
                                 lang["script"] = {}
-                            # use property_value_aliases to get ISO 15924 code
-                            if not script in lang["script"]:
-                                lang["script"][''.join(
-                                    unicodedataplus.property_value_aliases['script'][script]).lower()] = script.replace(
+                            #Uses property_value_aliases to get ISO-15924 code.
```
agutkin, line 82: #Uses -> # Uses

data/src/languages_update.py
cgibson6279: yup! I'm going to run `black -79` on everything before I leave today.

cgibson6279:
So I can confirm (as a biproduct of several foolish accidents as I was working last night) that `postprocess` runs A LOT faster if not updating `languages.json` during execution as implemented in my last push. I think it'd be best to do what @lfashby  suggested and have `postprocess` execute `languages_update.py ` first then run `split.py`. We're talking minutes as opposed to hours difference between the two implementations.

data/src/languages_update.py
```diff
@@ -10,14 +10,23 @@
...
+)
 
-from data.src.codes import LANGUAGES_PATH, TSV_DIRECTORY_PATH
+from data.src.codes import (
+    LANGUAGES_PATH,
+    TSV_DIRECTORY_PATH,
+)
 
 
 def _detect_best_script_name(
```
agutkin, line 26: An afterthought: Do you think whether the following may be more efficient:

1. Put all the characters of all the words in a single set.
2. Fill the mapping from characters to script names.
3. Pass the mapping to `_detect_best_script_name` and use the character as a key to get the script. This may, in theory, be more efficient than calling into `unicodedataplus`.

WDYT?

cgibson6279: In theory, it makes sense that it would be a bit more efficient, but I think we're talking about seconds saved, as opposed to minutes if we're updating outside of `split.py` as we are now. On my machine, it takes `languages_update.py` a minute or two max to run, so I would have to log the runtimes to see if we're actually more efficient at this point, but I think the suggested approach would definitely be better if we kept `_detect_best_script_name` in `split.py.`

kylebgorman:
I may have been unclear earlier: the bit "closes #319" goes in the PR description (the textbox), not the PR title. I don't know why this distinction exists but Github parses the description, not the title, for references to issues.

data/tsv/eng_latn_phonemic.tsv
```diff
@@ -1,5 +1,3 @@
-$deity	d e ɪ ɪ t i
```
lfashby, line 1: Based on a conversation with Cameron about `test_summary.py`, I noticed that I completely overlooked something. `eng_us_phonemic.tsv` has been renamed `eng_latn_phonemic.tsv`, the dialect information has been dropped. We've never had to handle a language having both dialect and script specifications and `split.py` is not setup to handle it (nor is [`_handle_wiki_name()`](https://github.com/kylebgorman/wikipron/blob/master/data/src/generate_tsv_summary.py#L17) in `generate_tsv_summary.py`).

This line overwrites the dialect information:
https://github.com/kylebgorman/wikipron/blob/c6c3b3f66dc2d85b0e0773000f8c7838fd40bf25/data/src/split.py#L48

This issue relates to #365 and I'm not really sure Cameron should have to fix it, but we'll need a fix for it before Cameron can run `generate_tsv_summary.py` and pass `test_summary.py`.

kylebgorman:
Good catch. Yes, it sounds like we need a fix for that ASAP. I think I'd
just fix it in #365 but feel free to make a separate issue if you want.

data/frequencies/README.md
```diff
@@ -4,7 +4,7 @@ Frequencies scripts
 The scripts in this directory are responsible for downloading word frequency
 counts from the [Leipzig Corpora
 Collection](https://wortschatz.uni-leipzig.de/en/download/) and merging those
-counts into [our corresponding TSVs](../tsv/).
+counts into [our corresponding TSVs](../../tests/tsv/).
```
kylebgorman, line 7: why is this `tests/tsv/`?

data/src/generate_tsv_summary.py
```diff
@@ -30,7 +30,15 @@ def _handle_wiki_name(
                     file_path,
                 )
                 continue
-            values = language[modifier][key]
+            # Temporary solution before issue 365 is resolved.
```
kylebgorman, line 33: Mark this "# TODO: remove temporary solution after #365."

data/src/languages_update.py
```diff
@@ -0,0 +1,163 @@
+#!/usr/bin/env python
+"""languages.json update
```
kylebgorman, line 2: Suggest: ```"""Updates languages.json.

...
```

File: data/src/languages_update.py [hunk 35]
```diff
@@ -0,0 +1,163 @@
...
+
+Takes languages.json file and updates the script entry for a language based on
+the files that are present in data/tsv.
+"""
```
kylebgorman, line 6: Should be one blank line after docstring though.

data/src/languages_update.py
```diff
@@ -0,0 +1,163 @@
...
+import collections
+import json
+import operator
+import os
```
kylebgorman, line 10: `unicodedataplus` is third-party so it goes after the typing imports but before the in-module imports.

data/src/languages_update.py
```diff
@@ -0,0 +1,163 @@
...
+import unicodedataplus  # type: ignore
+
+from typing import Dict, DefaultDict, Optional
+
+from data.src.codes import (  # type: ignore
```
kylebgorman, line 15: I think this can be one-line and if there's a typing problem it should be easy to fix since it's in-module.

data/src/split.py
```diff
@@ -60,10 +29,13 @@ def _iterate_through_file(
 
 def main() -> None:
     tsv_path = sys.argv[1]
```
kylebgorman, line 31: I realize this is not new but we really ought to be using argparse for this, just for consistency. If called without an argument it'll crash, but with an uninformative argument.

data/tsv/mww_latn_phonemic.tsv
```diff
@@ -1,17 +1,12 @@
 cag	c a ˧˩̤
-caj dab	c a ˥˧ d a ˥
```
kylebgorman, line 2: this seems to be removing every multi-word expression in this file.

kylebgorman (commented):
Few more comments here, plus a `black -l79` redo is needed when you're ready.

data/src/split.py
```diff
@@ -55,4 +56,15 @@ def main() -> None:
...
-    main()
+    parser = argparse.ArgumentParser(
+        description="Sets path to tsv",
+    )
+    parser.add_argument(
+        "-tsv_path",
+        type=str,
+        metavar="",
+        required=True,
+        help="Declare path to tsv files",
```
kylebgorman, line 67: This is somewhat surprising but the autogenerated help strings begin wtih lowercase so the ones I write do too. And TSV is an acronym so it should be capitalized in documentation: "path to TSV files" would then suffice.

data/src/languages_update.py
```diff
@@ -77,9 +79,9 @@ def _remove_mismatch_ids(
             str,
         ],
     ]
-) -> Dict[str, Dict[str, str,]]:
-    """If a values in lang["script"] appears more than once, the [key:value] pair that does not conform to ISO unicode
-    entries returned from unicodedataplus.property_value_aliases["script"] the key is deleted.
+) -> Dict[str, Dict[str, str, ]]:
```
kylebgorman: Remove the trailing comma and space here: `Dict[str, Dict[str, str]]:`

data/src/split.py
kylebgorman: Having not thought heavily about it I think if it runs reasonably fast (minutes, let's say) the design is solid.

agutkin (changes requested):
As I mentioned earlier, I think we need some logic to prevent the following:

data/tsv/cop_phonemic.tsv → data/tsv/cop_coptqaac_phonemic.tsv

cgibson6279:
@agutkin I'll work on adding that now

cgibson6279:
@agutkin so are we saying that we don't want 'zinh' and 'qaai' to be filtered out of the language scripts? The issue with coptic was that the alias returned by unicodedataplys was 'coptqaac' for some reason. But because inherited is really just diacritic type markers from what I understand, do we want those in the main scripts? At present they are filtered into another file and removed.

agutkin:
> @agutkin so are we saying that we don't want 'zinh' and 'qaai' to be filtered out of the language scripts? The issue with coptic was that the alias returned by unicodedataplys was 'coptqaac' for some reason. But because inherited is really just diacritic type markers from what I understand, do we want those in the main scripts? At present they are filtered into another file and removed.

Sorry for the confusion - Functionally, the change is perfectly fine. It's just the fact that we're keeping this strange compound as the name of the script in languages.json seems strange to me.

cgibson6279:
@agutkin gotcha. that makes sense, and I can add a fix for that quickly. It looks like it should only be 'copt' and 'zinh'  are the only two the return this dual alias if I'm looking correctly?

agutkin:
I hope so, but in the long run we may need a more robust solution. @kylebgorman, @lfashby WDYT?

data/tsv/mww_latn_phonemic.tsv
lfashby: Ah, we should have anticipated this. Any word with spaces in it (which will include a lot of Vietnamese entries) will get filtered out now.

(On a side note, are we supposed to be including words with spaces in them by default now? If not then the above entry in mww shouldn't be there and we need to rework our skip spaces logic.)

kylebgorman:
I think space-skipping is the default if I'm reading cli.py correctly.

lfashby:
Other than figuring out what to do with entries containing spaces this seems ready to go for me. Perhaps we should just bite the bullet and modify `_generalized_check()` in `split.py` to not filter out entries that are a mix of a single script and Common? Something like that.

Regarding Inherited and Coptic, those are the only two I’ve found that use multiple aliases so I think our current solution should be alright. It looks like ‘Qaac’ and ‘Qaai’ are retained as aliases for backwards compatibility and maybe in the future there will be more Q-type entries if ISO 15924 is updated?

kylebgorman (commented):
> Other than figuring out what to do with entries containing spaces this seems ready to go for me. Perhaps we should just bite the bullet and modify `_generalized_check()` in `split.py` to not filter out entries that are a mix of a single script and Common? Something like that.

I like that it's getting rid of `$deity` and other weird junk. Maybe we could have a special case for space, whereby it doesn't count?

data/tsv/haw_latn_phonemic.tsv
```diff
@@ -4,14 +4,12 @@ ahiahi	a h i a h i
 ako	a k o
 ala	a l a
 alana	a l a n a
-alapaʻi	a l a p a ʔ i
```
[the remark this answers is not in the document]
cgibson6279: @kylebgorman @lfashby I wonder if the new filtering might have something to do with the changes I made to `test_split.py`, particularly with the Inherited case?

cgibson6279:
@kylebgorman @lfashby I wonder if the new filtering might have something to do with the changes I made to test_split.py, particularly with the Inherited case?

data/src/common_characters.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python
+"""Creates json file containing list of "Common" for each language
+
+This module takes creats a json file for each language TSV file in data/tsv
```
agutkin, line 4: Typo: s/takes creats/creates

data/src/common_characters.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python
+"""Creates json file containing list of "Common" for each language
```
kylebgorman, line 2: "JSON". It's an acronym, like "TSV".

data/src/common_characters.py
```diff
@@ -0,0 +1,67 @@
...
+
+This module takes TSV files from data and return a json file for 
+each language TSV file in data/tsv that lists the unicode
```
kylebgorman, line 5: Unicdoe is a proper noun, so "Unicode".

data/src/common_characters.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python
+"""Creates json file containing list of "Common" for each language
+
+This module takes TSV files from data and return a json file for
```
kylebgorman, line 4: "JSON" again. 

"return" is a weird word here, I'd say it "writes" it or something like that.

data/src/common_characters.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python
+"""Creates json file containing list of "Common" for each language
+
+This module takes TSV files from data and return a json file for 
+each language TSV file in data/tsv that lists the unicode 
+typed "Common" characters that appear ineach language TSV file.
+"""
+import os
```
kylebgorman, line 8: Order should be:

One blank line after the docstring, then:

```
import os
import json
import unicodedata

from typing import Dict, Optional

import regex  # type: ignore
import unicodedataplus  # type: ignore

from codes import TSV_DIRECTORY_PATH
```

note:

* lexicographic sorting of `Dict` before `Optional`
* got rid of the type: ignore on `codes`...shouldn't be necessary?

data/src/languages_update.py
```diff
@@ -0,0 +1,171 @@
+#!/usr/bin/env python
+"""Updates languages.json
+
+This module takes the data/tsv directory as input and returns an
+updated version of languages.json, where the script entry for each
+language is updated to include every orthographic script present
+in each language's tsv file. Script entries are also updated
```
kylebgorman, line 7: "TSV"

data/src/languages_update.py
```diff
@@ -0,0 +1,171 @@
...
+to reflect such that script key entries match ISO 15924 aliases.
+"""
+
+import collections
+import json
+import operator
+import os
+
+from typing import Dict, DefaultDict, Optional
```
kylebgorman, line 16: `DefaultDict` before `Dict`.

data/src/languages_update.py
```diff
@@ -0,0 +1,171 @@
...
+
+import unicodedataplus  # type: ignore
+
+from data.src.codes import LANGUAGES_PATH, TSV_DIRECTORY_PATH  # type: ignore
```
kylebgorman, line 20: Surprised that this needs `type: ignore` since those are both obviously strings.

data/src/languages_update.py
```diff
@@ -0,0 +1,171 @@
...
+    and returns the ISO 15924 code alias for that script.
+
+    Example: "Arabic" -> "arab"
+    """
+    script = "".join(
+        unicodedataplus.property_value_aliases["script"][value]
+    ).lower()
+    # Removes qaac tag from end of script.
+    script = script.replace("qaac", "")
+    script = script.replace("qaai", "")
```
kylebgorman, line 74: do `return script.replace("qaai", "")` here and cut lines 75-76.

data/src/languages_update.py
```diff
@@ -0,0 +1,171 @@
+#!/usr/bin/env python
```
kylebgorman, line 1: Please run this whole file through `black -l79` again.

data/src/postprocess
```diff
@@ -12,4 +13,4 @@ main() {
     done
 }
 
-main
+main
```
kylebgorman, line 16: Put a newline at the end of this plesae.

data/src/split.py
```diff
@@ -1,15 +1,10 @@
 #!/usr/bin/env python
 
-import collections
 import json
```
kylebgorman, line 3:
```import argparse
import json
import os
...
``` etc.

File: data/tsv/bel_cyrl_phonetic.tsv [hunk 55]
```diff
@@ -1,5 +1,3 @@
-аб'явіць	a b j a vʲ i t͡sʲ
```
kylebgorman, line 1: The language selector should not be throwing these out. Apostrophe is part of the Belarusian orthography. It's used to separate a hard (non-palatalized) consonant from the following iotated vowel (otherwise it'd be soft), thus в'етнам is [vjetnam] 'Vietnam and not [vʲjetnam]. Is there some way to modify it to not exclude such words?

data/tsv/bdq_latn_phonemic.tsv
```diff
@@ -10,7 +10,6 @@ blok	b l ɔː k
 blu	b l uː
 băt	b a t
 bơbe	b ə b ɛː
-bơbư̆ng	b ə b ɨ ŋ
```
kylebgorman, line 13: Do we know these words should be excluded?

data/tsv/bre_latn_phonemic.tsv
```diff
@@ -23,8 +23,6 @@ aour	ɔ w r
 arall	a r a lː
 aramaeg	a r a m a ɛ k
 arar	a ʁ a ʁ
-arc'hanti	a ʁ χ ɑ̃ n t i
```
kylebgorman, line 26: I don't know anything about Breton spelling but it looks like c'h is just how you spell dorsal fricatives of various types? These should probably not be filtered.

data/tsv/bur_mymr_phonemic.tsv
```diff
@@ -1425,7 +1425,6 @@
 ထမင်းချဉ်	tʰ ə m ɪ́ ɴ d͡ʑ ɪ̀ ɴ
 ထမင်းဆိုင်	tʰ ə m ɪ́ ɴ z à ɪ ɴ
 ထမင်းသုပ်	tʰ a̰ m ɪ́ ɴ θ o ʊ ʔ
-ထမင်း‌ကြော်	tʰ ə m ɪ́ ɴ‌ d͡ʑ ɔ̀
```
kylebgorman, line 1428: I don't understand why these words are being excluded.

data/tsv/cat_latn_phonemic.tsv
```diff
@@ -3965,55 +3965,6 @@ alvocats	ə l b u k a t s
 alvocats	ə l v o k a t s
 alzina	a l z i n a
 alzina	ə l z i n ə
-al·legada	a l l e ɡ a d a
```
kylebgorman, line 3968: These words should not be excluded:

https://en.wikipedia.org/wiki/Catalan_orthography#Punt_volat_(middot)

data/tsv/cat_latn_phonemic.tsv
```diff
@@ -16672,16 +16464,6 @@ cúrcuma	k u ɾ k u m a
 cúrcuma	k u ɾ k u m ə
 cúter	k u t e ɾ
 cúter	k u t ə
-d'	d
```
kylebgorman, line 16675: nor these

kylebgorman (commented):
Suggestions at this stage:

1. Please handle any in-line comments to code and resolve them when done.
2. You regenerated (with new line endings?) files in `data/phones`. Since your change has nothing to do with that, could you just revert those changes from the CL altogether? Should make it easier to review.
3. Same for the `requirements.txt` file.
4. In languages like Belarusian, Breton, and Catalan, the current rules are too restrictive. Belarusian and Catalan both use an apostrophe character (whether or not it's the ASCII one, I don't know) for various things, and Catalan uses a mid-dot between a <ll> when that's not the palatal fricative, but those entries are all being thrown out. The rules for how we classify scripts need to permit those characters (among others) to be used anywhere...

data/tsv/bur_mymr_phonemic.tsv
agutkin: This gets excluded because of the ZWJ/ZWNJ which are common to all Brahmic scripts. We need solution for this as well.

```
['MYANMAR LETTER THA', 'MYANMAR LETTER MA', 'MYANMAR LETTER NGA', 'MYANMAR SIGN ASAT', 'MYANMAR SIGN VISARGA', 'ZERO WIDTH NON-JOINER', 'MYANMAR LETTER KA', 'MYANMAR CONSONANT SIGN MEDIAL RA', 'MYANMAR VOWEL SIGN E']
```

data/tsv/bdq_latn_phonemic.tsv
agutkin: I suppose the combining breve is allowed for whichever language this is, isn't it?

```
['LATIN SMALL LETTER B', 'LATIN SMALL LETTER O WITH HORN', 'LATIN SMALL LETTER B', 'LATIN SMALL LETTER U WITH HORN', 'COMBINING BREVE', 'LATIN SMALL LETTER N', 'LATIN SMALL LETTER G']
```

data/tsv/bur_mymr_phonemic.tsv
lfashby: I suppose a quick (and temporary) solution would be to extend our `regex_string` to include `\u200C\u200D`? I think in the future we'll only want to extend `regex_string`'s coverage uniquely for each language but extending it for all languages now is a bit easier.

kylebgorman:
Some notes on what to keep:

1. APOSTROPHE is in, as are variants like RIGHT SINGLE QUOTATION MARK, MODIFIER LETTER APOSTROPHE, and LEFT SINGLE QUOTATION MARK. 
2. A bunch of languages use ZERO WIDTH SPACE and ZERO WIDTH NON JOINER. Let's keep that for now but make a note to study how they're used.
3. MIDDLE DOT is in, it's part of Catalan amongst others.
4. Arabic and Katakana specials are in.
5. Tilde is in (Phalura appears to use it as a ASCII character for nasalization, who knew)
6. There's a bunch of weird stuff only used in Cantonese. Let's remove those and later we can ask Jackson to tell us what the biz is.
7. All other ASCII punctuation is out.
8. It looks to me like everything under "inherited" is legit.

If you could implement this and regenerate, then I can spot check some more.

lfashby:
4 and 8 should be easy. 

The Katakana and Hiragana specials should already pass through and Arabic can be covered by modifying `prop` in `_generalized_check` (in addition to the Inherited changes below):
```
    prop = (
        "Block" if script == "Katakana" or script == "Hiragana" or script == "Arabic" else "Script"
    )
```
Allowing all inherited characters to pass can be accomplished via adding `\p{{Script=Inherited}}` to `regex_string`.

cgibson6279:
Lucas mentioned that it may be computationally expensive to have the `split.py` have to reconstruct the regex string every time it's run in `postprocess` so I'm working on devising a better way to handle that if possible.

data/tsv/egy_latn_phonemic.tsv
```diff
@@ -3331,7 +3263,6 @@ zꜣṯw	z ɑ t͡ʃ uː
 ꜥnḫ	ʕ aː n a χ
 ꜥnḫ	ʕ aː n ə χ
 ꜥnḫ	ʕ oː n ə χ
-ꜥnḫ.tj	ɑː n x ʔ t i
```
agutkin, line 3334: Intresting. I don't know much about transliteration schemes for Ancient Egyptian, but "." may indeed stand for something reasonably valid. How do we check?

data/tsv/lat_latn_vulg_phonemic.tsv
```diff
@@ -1,536 +1,3 @@
-*abanteō	a β a n t e oː
-*abbatuō	a β β a t u oː
-*abbatō	a β β a t oː
-*abbiberō	a β β i β e r oː
-*absedium	a p s ɛ d i u m
-*absediī	a p s ɛ d i iː
-*abyssimus	a β e s s i m u s
```
agutkin, line 7: Prefix asterisks in Latin entries (here and in the Ecclesiastical version) refer to reconstructed words. Again, not clear what's the way going forward with those.

kylebgorman (commented):
Good news, I just spot-checked every language affected by this and the only remaining issue I see is what to do with Egyptian. Before submitting:

1. Figure out some solution to Egyptian.
2. Remove extraneous files added in this PR.
3. Revert all changes to the `phones` directory.
4. Handle any style concerns.
5. Re-run `black`.

data/tsv/lat_latn_vulg_phonemic.tsv
kylebgorman: I wouldn't say that's WAI exactly but it's a nice side-effect. Please keep it that way!

kylebgorman:
> Lucas mentioned that it may be computationally expensive to have the `split.py` have to reconstruct the regex string every time it's run in `postprocess` so I'm working on devising a better way to handle that if possible.

I was wondering about that too. I couldn't tell from the way things were being passed around whether that was necessary or if it could just be a constant.

data/src/common_characters.py
```diff
@@ -0,0 +1,117 @@
...
+each language TSV file in data/tsv that lists the Unicode
+typed "Common" and "Inherited" characters that appear
+in each language TSV file -
+common_char_summary_by_lang.
+
+It also writes a JSON file that includes a full
+"Common" and "Inherited" character set for all files -
+global_common_char_summary.json.
+"""
+import os
```
kylebgorman, line 14: Not in lexicographic order here.

data/src/common_characters.py
```diff
@@ -0,0 +1,161 @@
...
+                        ][inh_char_name] = inh_char
+        with open(COMMMON_CHAR_LIST_PATH, "w", encoding="utf-8") as write_path:
+            json.dump(master_set, write_path, ensure_ascii=False, indent=4)
+    # Creates global master common/inherited set.
+    global_set: Dict[str, Dict[str, str]] = {}
+    global_set["Common"] = {}
+    global_set["Inherited"] = {}
+    for symbol in master_set.values():
+        for char_type, char_symbol in symbol.items():
+            if char_type == "Common":
```
kylebgorman, line 146: Suggested simplification:

```
for char_type, char_symbol in symbol.items():
    if char_type in ("Common", "Inherited"):
       global_set[char_type].update(char_symbol)
```

kylebgorman (approved):
LGTM

PR #367, opened 2021-03-07
[geo] Adds CG. Fixes errors in phonelist.

ajmalanoski:

As I was making the covering grammar for Georgian (@Othergreengrasses), I noticed there were some misleading aspects to the Georgian phonelist, namely, the inclusion of <s‘> and <ʃ‘> without acknowledging that these are not phonemes of Georgian, but rather used to transcribe the ejective affricates (or so I assumed).

However, it turns out the affricates are transcribed with tie bars on Wiktionary, seemingly fairly consistently (at least based on the handful I checked). Yet the data for Georgian (both filtered and unfiltered) has no tie bars in it whatsoever. For example, [ჩაფი](https://en.wiktionary.org/wiki/%E1%83%A9%E1%83%90%E1%83%A4%E1%83%98) seems to have had a tie bar in its pronunciation for [as long as it's had a pronunciation on Wiktionary](https://en.wiktionary.org/w/index.php?title=%E1%83%A9%E1%83%90%E1%83%A4%E1%83%98&type=revision&diff=58547541&oldid=36845782), yet is transcribed without a tie bar in our data. As far as I know, `/data/src/scrape.py` doesn't remove tie bars, so this is puzzling.


ajmalanoski:
Oops, I forgot to do the postprocessing after editing the phone list. I'll take care of that now.

ajmalanoski:
Hmm… I'm getting some errors that I think are related to my recent switch to conda, although I'm not sure. I'm out of time to work on this tonight, so I'll try to figure it out tomorrow

lfashby:
This is very confusing. I just scraped Georgian and found that the tie bars were present in the scraped data, so that's good. However, I can't think of any reason why the tie bars were absent in our recent Georgian data and in all previous scraped Georgian data. I even reset my repo back to a commit from a big scrape from last year (in which no tie bars were present in the Georgian data) and ran a Georgian scrape (making use of `cut_off_date`) but still got tie bars.

I'm not sure what's going on. `segments` never had any problems with tie bars and even if it did it wouldn't only affect Georgian. Perhaps I'm missing something obvious. Or perhaps our Georgian data is haunted!

ajmalanoski:
It looks like it can't find WikiPron (even though I'm running the code from inside WikiPron). Any idea why?
```
Traceback (most recent call last):
  File "./scrape.py", line 13, in <module>
    import wikipron  # type: ignore
ModuleNotFoundError: No module named 'wikipron'
```

lfashby:
> It looks like it can't find WikiPron (even though I'm running the code from inside WikiPron). Any idea why?
> 
> ```
> Traceback (most recent call last):
>   File "./scrape.py", line 13, in <module>
>     import wikipron  # type: ignore
> ModuleNotFoundError: No module named 'wikipron'
> ```

If you type `which wikipron` from the terminal what do you see? Are you on a different computer and have you not [installed](https://github.com/kylebgorman/wikipron#development) wikipron?

ajmalanoski:
Oh, I forgot to run `pip install --no-deps -e .` after switching to conda. It looks like that fixed it :) Thanks!

data/phones/geo_phonemic.phones
```diff
@@ -1,11 +1,11 @@
 # Based on https://en.wikipedia.org/wiki/Help:IPA/Georgian
-# Vowels.
+# VOWELS
```
kylebgorman, line 2: Why is this shouting at me? ;)

kylebgorman (approved):
LGTM.

data/phones/geo_phonemic.phones
ajmalanoski: It's not shouting, it's just excited!

ajmalanoski:
Do you have any other comments, @lfashby? Otherwise, I'll go ahead and merge

lfashby (approved):
I don't have any other comments. I suppose we'll have to leave the case of the missing tie bars unsolved and just re-scrape Georgian.

ajmalanoski:
I re-scraped it for this commit, so that's already done. We just need to move the new data over to the repo for the task, I think. In any event, I'll squash and merge.

PR #369, opened 2021-03-13
Morphology stub

kylebgorman:

### DIFF
--- a/data/morphology/unimorph_languages.json
+++ b/data/morphology/unimorph_languages.json
@@ -1,0 +1,4 @@
+{
+    "eng": "http://raw.githubusercontent.com/unimorph/eng/master/eng",
+    "ger": "http://raw.githubusercontent.com/unimorph/deu/master/deu"
+}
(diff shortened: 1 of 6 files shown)


data/morphology/unimorph_languages.json
```diff
@@ -0,0 +1,4 @@
+{
```
reubenraff, line 1: we want the raw github links with the lgs as the only wikipron lg code abbreviation or do we want the format to look like the JSON file format like this example for armenian? 


"hye-am_web_2017_1M": {
        "path": [
            "../tsv/arm_e",
            "../tsv/arm_w"
        ],
        "data_url": "https://raw.githubusercontent.com/reubenraff/wikipron/master/data/tsv/arm_e_phonetic.tsv"
    },

kylebgorman: We don't need any of that additional metadata here, it seems to me, yet. (UniMorph doesn't distinguish between Western and Eastern Armenian.) We just need the URL of the matching UniMorph file and the language code.

reubenraff: ok I'll format as such in the unimorph_data JSON file

reubenraff (approved):

PR #371, opened 2021-03-18
adds data files from directory `unimorph_data`

reubenraff:
git had an error saying the finnish data files are too large to append to this commit. Should I divide them into 2 files each ?


kylebgorman (commented):
1. It's not my intention that we check in those data files, ever, since as you see, they're quite large.  Can you remove them from the PR? The user who needs them can just run the script.
2. It looks to me like you didn't check in changes to `unimorph_languages.json`, which is the most important part here. Can you include those?
3. It looks like your `get_unimorph_data.py` replicates the functionality of  `grab_unimorph_data.py` added in #369. Can you remove yours, and if it has any additional features, integrate its functionality into the existing program?

kylebgorman:
I see the `black` test is failing. You can fix that just by reflowing the code (running `black -l79 *.py` in the morphology directory), probably.

reubenraff:
> I see the `black` test is failing. You can fix that just by reflowing the code (running `black -l79 *.py` in the morphology directory), probably.

sgtm

kylebgorman:
Would you like me to close this since you're working on #372, I assume as a replacement?

kylebgorman:
Closing from inactivity; happy to discuss whether we actually want to check these files in at some point.

PR #376, opened 2021-03-20
Uses one-to-many for UniMorph downloading

### LINKED ISSUE #373: JSON UniMorph morphology: make one-to-many
Once #372 by @reubenraff is submitted, we should expand the JSON file and associated logic so that the mapping from WikiPron language codes to UniMorph URLs is one-to-many. This will allow us to deal with the fact that `fin` (Finnish) is two files.

1. Instead of `Dict[str, str]` make the UniMorph JSON a `Dict[str, List[str]]` instead. Most languages will only have one entry in the list.
2. In the `grab_unimorph_data.py` script, loop over the list of URLs for each language, writing all of them into the WikiPron language code + `.tsv`. That'll give us a single `fin.tsv` file (for instance).

We could make it the case that the dictionary values are polymorphic (`Union[str, List[str]]`) but I think that'd just make things more confusing.

kylebgorman:
The Finnish file is split into two because of GitHub file size
limitations. To get around this we make the value portion of the JSON
dictionary a list.

Closes #373.


lfashby (approved):
LGTM just one comment.

data/morphology/grab_unimorph_data.py
```diff
@@ -6,30 +6,32 @@
...
-                logging.info(
-                    "Status code %d while downloading %s",
-                    response.status_code,
-                    language,
-                )
-                to_retry[language] = data_to_grab[language]
+    for language, urls in data_to_grab.items():
+        with open(f"tsv/{language}.tsv", "wb") as sink:
+            for url in urls:
+                with requests.get(url, stream=True) as response:
```
lfashby, line 23: I think this might borrow too much from the `grab_wortschatz_data.py` script. All of the checking of status codes and `stream=True`, `to_retry` and `time.sleep` stuff was because downloading the Wortschatz data often failed in exotic ways. I doubt we need any of it here, but better safe than sorry?

kylebgorman: I see why you say that but FWIW a few of these files are on the order of 100 MB and it's easy to imagine it hanging up with something that big. I figured since it ain't broke...

PR #387, opened 2021-03-23
Final newlines for all *.tsv and *.py files.

### LINKED ISSUE #386: Final newlines
GitHub seems to be goofy about final newlines these days. Maybe we should audit our data files for their presence.

kylebgorman:
Closes #386.

Script used to test for this:

```bash
for FILE in $@; do
    if [[ ! -d "${FILE}" ]]; then
        if [[ $(tail -c1 "${FILE}" | wc -l) -eq 0 ]]; then
            echo "Missing newline at eof: ${FILE}"
        fi
    fi
done
```


data/src/generate_phones_summary.py
```diff
@@ -53,7 +53,8 @@ def main() -> None:
         if file_path.endswith(".md") or file_path.endswith("tsv"):
             continue
         with open(f"{path}/{file_path}", "r", encoding="utf-8") as phone_list:
-            num_of_entries = sum(1 for line in phone_list)
+            # We subtract one for the final newline.
+            num_of_entries = sum(1 for line in phone_list) - 1
```
ajmalanoski, line 57: I just noticed this, but this script counts comments as "entries" (which isn't what we want, I should think). I guess the line should be `num_of_entries = sum(1 for line in phone_list if line[0] != "#") - 1` or something to that effect.

ajmalanoski:
I don't really know Bash, so I can't comment on your script. The edits themselves look fine though. My only concern is with how phones are being counted (as discussed in my comment on `generate_phones_summary.py`).

data/src/generate_phones_summary.py
kylebgorman: +1. I'll create an issue.

ajmalanoski:
My bad for not being more thorough earlier, but some of the `.phones` files still don't end in blank lines. Offenders include the Latin phone list and both Spanish phone lists (perhaps among others; I haven't gone through all of them yet)

kylebgorman:
Okay, I was having some weird troubles. Feel free to reopen.

ajmalanoski:
No worries. I was thinking about cleaning up the `.phones` files (in a lot of them, the phones aren't listed in any particular order, which makes them a bit hard to read IMO), so I can add the lines at the ends of the files if/when I do that

PR #391, opened 2021-03-24
Add new columns to summary tables

### LINKED ISSUE #365: New columns in the summary tables
This issue suggests some minor enhancements to the autogenerated tables in `data/src/README.md` and `data/src/tsv_summary.tsv`.

1. The column `Phonetic/Phonemic` currently has labels like `Phonetic_Filtered`. These are two independent pieces of information which are not closely related. A new column "Filtered" (with True and False values) should be introduced instead.
2. Arguably, dialect information should be in a separate column than language information. For languages without a dialect specification the column would just be blank.
3. PR #364, which addresses issue #319, adds script information to every file. This should go in a separate column from language, dialect, etc.

Example entry under the proposed:

```
TSV | wel | Welsh | North Wales | Latin | True | Phonemic | True | 8,477 
```

blueCormorant:
Closes #365.


data/src/generate_tsv_summary.py
```diff
@@ -11,15 +11,60 @@
...
 
+def _get_modifiers(
+    language: Dict[str, Any], modifier: str
+) -> Dict:
+    try:
+        return language[modifier]
+    except KeyError:
+        return {}
+
+def _handle_script(
```
kylebgorman, line 24: The logic in these `_handle_*` functions is almost identical. Could we not just have a single one?

data/src/generate_tsv_summary.py
```diff
@@ -11,15 +11,60 @@
...
+    language: Dict[str, Any], file_path: str
+) -> Dict:
+    scripts = _get_modifiers(language, "script")
+    key = file_path[
+        file_path.index("_") + 1 : file_path.rindex("_phone")
+    ]
+    if "_" in key:
+        key = key[:key.index("_")]
+    if key in scripts:
```
kylebgorman, line 33: This is faster than what you have in lines 33-36.

```
return scripts.get(key, "")
```

Why? It only looks up once (whereas yours looks the key up in 33, and then again in line 34).

Also applies to some of the other _handle functions.

data/src/generate_tsv_summary.py
```diff
@@ -69,18 +114,20 @@ def main() -> None:
...
         iso639_code = file_path[: file_path.index("_")]
-        transcription_level = file_path[
-            file_path.index("phone") : file_path.index(".")
-        ].capitalize()
-        wiki_name = _handle_wiki_name(
-            languages[iso639_code], file_path, modifiers
-        )
+        transcription_level = _handle_transcription_level(file_path)
+        wiki_name = languages[iso639_code]["wiktionary_name"]
+        filtered = True if "filtered" in file_path else False
```
kylebgorman, line 119: Instead of this you can write:

```filtered = "filtered" in file_path```

The True/False wrapping here is unnecessary.

**Review by kylebgorman — COMMENTED** [review]
This basically looks good to me; one or two minor style notes. I'll ask @lfashby to take a quick look when he has a minute too.

I think you'll need to run `black -l79` and check `flake8` on the affected *.py files before submitting.

**Comment by kylebgorman** [comment]
Please edit the description so that it includes the string:

```Closes #365.```

GitHub will then close the issue and cross-link it when this PR is merged.

File: CHANGELOG.md [hunk 4]
```diff
@@ -13,7 +13,7 @@ Unreleased
 ### Under `data/`
 
 #### Added
-
+`
```
kylebgorman, line 16: I assume you didn't intend this?

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ Unreleased
 
 #### Changed
 
+-	Added new columns in summary tables (\#365)
```
kylebgorman, line 27: Align with the rest of the text (this is more indented) and add sentence punctuation (i.e., a period at teh end) like the other entries.

data/src/generate_tsv_summary.py
```diff
@@ -11,15 +11,46 @@
 
 
 def _wiki_name_and_transcription_level(ele: List[str]) -> str:
-    return ele[3] + ele[5]
+    return ele[3] + ele[7]
 
+def _get_modifiers(
+    language: Dict[str, Any], modifier: str
+) -> Dict:
+    try:
```
kylebgorman, line 19: This could also be:

```return language.get(modifier, {})```

for lines 19-22.

File: CHANGELOG.md (hunk 4 above)
Comment (blueCormorant) [reply CHANGELOG.md:16]: Yeah that was a mistake.

**Comment by blueCormorant** [comment]
Hey black and flake8 don't seem to agree with each other with regard to whitespace usage. I made sure to run black -l79.

I get `whitespace before ':'` on lines like `file_path.index("phone") : file_path.index(".")`

**Comment by kylebgorman** [comment]
> Hey black and flake8 don't seem to agree with each other with regard to whitespace usage. I made sure to run black -l79.
> 
> I get `whitespace before ':'` on lines like `file_path.index("phone") : file_path.index(".")`

Yes, this is annoying. Remove the whitespace after `black`-ing the code and I think the tests will pass, and if not, we can submit even though they don't. (Note you also have a `mypy` issue: your helper functions are not returning what they claim to be.)

File: data/src/generate_tsv_summary.py [hunk 7]
```diff
@@ -11,15 +11,46 @@
 
 
 def _wiki_name_and_transcription_level(ele: List[str]) -> str:
-    return ele[3] + ele[5]
+    return ele[3] + ele[7]
 
+def _get_modifiers(
+    language: Dict[str, Any], modifier: str
+) -> Dict:
```
kylebgorman: What kind of dict? Usually one declares the key and value.

data/src/generate_tsv_summary.py
```diff
@@ -11,15 +11,46 @@
...
+    try:
+        return language[modifier]
+    except KeyError:
+        return {}
+
+def _handle_modifier(
+    language: Dict[str, Any], file_path: str, modifier: str,
+) -> Dict:
```
kylebgorman, line 21: Ditto.

data/src/generate_tsv_summary.py
```diff
@@ -97,17 +124,29 @@ def main() -> None:
...
             "| Link | ISO 639-2 Code | ISO 639 Language Name "
-            "| Wiktionary Language Name | Case-folding "
-            "| Phonetic/Phonemic | # of entries |",
+            "| Wiktionary Language Name | Script | Dialect | Filtered "
+            "| Phonetic/Phonemic | Case-folding | # of entries |",
             file=sink,
         )
         print(
-            "| :---- | :----: | :----: | :----: | :----: | :----: | ----: |",
+            "| :---- |" + "".join([" :----: |" * 8]) + " ----: |",
```
kylebgorman, line 132: Sorry for late-breaking observation but I just noticed this. What you're doing here, I think, is conceptually:

```
"".join([str])
```

You can just write:

```str```

instead. That would look like:

````"| :---- |" + " :----: |" * 8 + " ----: |",```

I think.

**Review by kylebgorman — COMMENTED** [review]
Basically LGTM. One last thing, and if tests pass...

**Review by lfashby — COMMENTED** [review]
Looking good. Just two comments below, one of which is possibly irrelevant.

#389 will require making some changes to this file again but I think we've got all the right pieces in place.

File: data/src/generate_tsv_summary.py [hunk 10]
```diff
@@ -11,7 +11,31 @@
...
+def _handle_transcription_level(file_path: str) -> str:
+    trans = file_path[
+        file_path.index("phone"):file_path.index(".")
+    ].capitalize()
+    if "_" in trans:
+        trans = trans[: trans.index("_")]
+    return trans
 
 
 def _handle_wiki_name(
```
lfashby, line 41: We can dump this entire function now right? However, you'll want to integrate the
```
            if "|" in values:
                values = values.replace(" |", ",")
```
functionality somewhere else since our English dialects are stored like so:
```
        "dialect": {
            "us": "US | General American",
            "uk": "UK | Received Pronunciation"
        },
```
And that "|" is throwing off the README table for English.

data/src/generate_tsv_summary.py
```diff
@@ -11,7 +11,31 @@
 
 
 def _wiki_name_and_transcription_level(ele: List[str]) -> str:
-    return ele[3] + ele[5]
+    return ele[3] + ele[7]
+
+
+def _handle_modifier(
```
lfashby, line 17: Now that all files have scripts I guess we could grab the script more directly and just call this once, something like:
```
modifiers = language.get("dialect", {})
start = file_path.index("_") + 1
script_key = file_path[start:file_path.index("_", start)]
dialect_key = file_path[file_path.index("_", start) + 1: file_path.rindex("_")]
return language['script'][script_key], modifiers.get(dialect_key, "")
```
I'm not sure whether that's a worthwhile improvement though.

data/src/generate_tsv_summary.py
kylebgorman: Good catch. I guess replace " | " with ", " maybe?

kylebgorman: +1

data/src/generate_tsv_summary.py
kylebgorman: +1.

blueCormorant:
What do you think of putting "N/A" instead of the empty string as a place filler for scripts that do not have a dialect? I think it would make the raw summary table easier to read.

kylebgorman:
I don't love it personally (feel like empty string is fine) but will defer
to Lucas for a tie-breaking vote.

lfashby:
Do you mean the `tsv_summary.tsv` table? I think only the `README.md` needs to be (human) readable and the [formatting](https://github.com/kylebgorman/wikipron/blob/6282c111f0d9a21102c568e479099580929f366f/data/README.md) it is given when viewed on Github (which is ideally how people interact with it) makes it look okay to me.

kylebgorman:
I think in the README this would make sense, but would be weird in the TSV
file.

kylebgorman (approved):
LGTM. Hit the big green button at the bottom when you're ready.

PR #395, opened 2021-03-26
Reorganization of `phones`

ajmalanoski:
- [x] Updated `Unreleased` in `CHANGELOG.md` to reflect the changes in code or data.

This PR implements some of the organization suggestions for the `data/phones` directory suggested in #360 . Changes include the following:

- created new directories `data/phones/lib` and `data/phones/phones`
- moved `.phones` files from `data/phones` to `data/phones/phones` 
- moved `generate_phones_summary.py`, `list_phones.py`, and `normalize.py` from `data/src` to `data/phones/lib`
- updated paths in `data/scrape/lib/codes.py`, `data/phones/HOWTO.md` and `tests/test_data/test_summary.py`

I did not make a file `data/phones/postprocess` because I don't know enough Bash 🤷 Also, I'm not sure what would go in that file anyway

PR #394, opened 2021-03-25
Partial data side reorganization

lfashby:
- [x] Updated `Unreleased` in `CHANGELOG.md` to reflect the changes in code or data.

This PR contains a partial reorganization of `wikipron/data` according to the proposal in #360. I moved all of the scraping type scripts into a new `data/scrape` directory, reworked some paths given those changes, added a temporary `data` README that we can fill out once `src` is fully removed, and fixed a few oversights in a couple scraping scripts.

PR #396, opened 2021-03-26
Reorganization of `cg`

### LINKED ISSUE #385: Covering grammar naming
As it stands, the covering grammar files in `data/cg/tsv` are just named `<lng>.tsv`, where `<lng>` is the three-letter language code. I think the file names should for the CGs should match those used for the data they correspond to, that is, they should phonetic/phonemic status and script name.

ajmalanoski:

Closes #385 

This PR changes the organization of the `data/cg` directory in a similar way to #394 and #395. Changes include the following:

- Renamed `data/cg` to `data/covering_grammar` and added folders `data/covering_grammar/lib` and `data/covering_grammar/tsv`
- Renamed covering grammar files to include script info and transcription type (phonemic/phonetic)
- Moved relevant scripts from `data/src` and `data/src/error_analysis` to `data/covering_grammar/lib`
- Added a file to make the input files for `data/covering_grammar/lib/error_analysis.py` (for some reason, this file wasn't in WikiPron, even though it existed at the time that Arundhati added the CG stuff)
- Added a placeholder README for `data/covering_grammar`

I imagine we'll eventually want the README to be a table like the ones we have for `data/phones` and `data/scrape`, and that we'll probably want some sort of `generate_cg_summary.py` to make that table.


data/covering_grammar/lib/make_test_file.py
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env python
+
```
kylebgorman, line 2: While I acknowledge you're right that you have a blank line here, let's standardize without it?

data/covering_grammar/lib/make_test_file.py
```diff
@@ -0,0 +1,45 @@
...
+import argparse
+import logging
+
+
+def main(args: argparse.Namespace) -> None:
+    with open(args.gold, "r") as gf, open(args.pred, "r") as pf:
+        with open(args.out, "w") as wf:
+            for i, (g_line, p_line) in enumerate(zip(gf, pf)):
+                try:
```
kylebgorman, line 18: This try/assert/except AssertionError design is a weird one for me. Instead of that, I'd do an if-statement, log the warning, and continue.

data/covering_grammar/lib/make_test_file.py
```diff
@@ -0,0 +1,45 @@
...
+                    g_data = g_line.split("\t")
+                    p_data = p_line.split("\t")
+                    # Make sure that gold data and predictions have the
```
kylebgorman, line 21: Please add sentential punctuation to these comments, and throughout.

data/covering_grammar/lib/make_test_file.py
```diff
@@ -0,0 +1,45 @@
...
+"""
+
+import argparse
+import logging
+
+
+def main(args: argparse.Namespace) -> None:
+    with open(args.gold, "r") as gf, open(args.pred, "r") as pf:
+        with open(args.out, "w") as wf:
+            for i, (g_line, p_line) in enumerate(zip(gf, pf)):
```
kylebgorman, line 17: This line number will be off by one because we start counting as zero. To fix I think you'll do:

`enumerate(zip(gf, pf), 1)`

Also while I'm bikeshedding, `i` is not a great name here. I'd call it `lineno` or something like that?

data/covering_grammar/lib/make_test_file.py
```diff
@@ -0,0 +1,45 @@
...
+                try:
+                    g_data = g_line.split("\t")
```
kylebgorman, line 19: Why don't you unpack these here? E.g., `(g_pron, g_word)`?

Indexing into a small tuple like you do later is code smell, to me.

data/covering_grammar/lib/make_test_file.py
```diff
@@ -0,0 +1,45 @@
...
+                    p_data = p_line.split("\t")
+                    # Make sure that gold data and predictions have the
+                    # same words
+                    assert g_data[0] == p_data[0]
+                    word = g_data[0]
+                    # Note that we use `strip` to remove the newline
+                    g_pron = g_data[1].strip()
+                    p_pron = p_data[1].strip()
+                    line = "\t".join([word, g_pron, p_pron])
```
kylebgorman, line 28: I'd just do `line = f"{word}\t{g_pron}\t{p_pron}"`; I think that's easier to read.

data/covering_grammar/lib/make_test_file.py
```diff
@@ -0,0 +1,45 @@
...
+                    print(line, file=wf)
+                except AssertionError:
+                    logging.warning(f"{g_data[0]} != {p_data[0]} (line {i})")
```
kylebgorman, line 31: Here I'd use old-style string interpolation:

`logging.warning("%s != %s (line %d)", g_word, p_word, i)`

data/covering_grammar/lib/make_test_file.py
```diff
@@ -0,0 +1,45 @@
...
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument(
+        "gold", help="TSV with words and correct pronunciations"
+    )
+    parser.add_argument(
+        "pred", help="TSV with words and predicted pronunciations"
+    )
+    parser.add_argument(
+        "-o", "--out", help="file to write to", default="out.tsv"
```
kylebgorman, line 43: I think having a default output file here is probably not a good idea; I'd make it mandatory.

`--out` is so short that I don't see any advantage to having a shortopt `-o`.

data/covering_grammar/lib/make_test_file.py
```diff
@@ -0,0 +1,45 @@
...
+                    g_pron = g_data[1].strip()
+                    p_pron = p_data[1].strip()
+                    line = "\t".join([word, g_pron, p_pron])
+                    print(line, file=wf)
+                except AssertionError:
+                    logging.warning(f"{g_data[0]} != {p_data[0]} (line {i})")
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description=__doc__)
```
kylebgorman, line 35: Don't you need to set `logging.basicConfig` here?

kylebgorman (commented):
Agree we may want a covering grammar generated README too, why not. It'd look cool and encourage more submitted CGs maybe.

The error analysis code and README is still here: `data/src/error_analysis`.

I just ruthlessly bikesheaded your `make_test_file.py`, sorry about that.

data/covering_grammar/lib/make_test_file.py
```diff
@@ -0,0 +1,45 @@
...
+each line has a word, its gold pronunciation, and the predicted pronunciation.
+Assumes that the input files have the same words in the same order.
+"""
+
+import argparse
+import logging
+
+
+def main(args: argparse.Namespace) -> None:
+    with open(args.gold, "r") as gf, open(args.pred, "r") as pf:
```
kylebgorman, line 15: I don't know if this is a plus here or not but when you're doing a lot of file-opening, contextlib's `ExitStack` helps a lot: https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack

lfashby:
I don’t have much to add. I generally prefer more verbose variable names but that's just me. After this we just have to write up these temporary TSVs.

> The error analysis code and README is still here: data/src/error_analysis.

Are you saying that those file should stay in `data/src`? They’ve moved to `data/covering_grammar/lib` from what I can see.

kylebgorman:
> > The error analysis code and README is still here: data/src/error_analysis.
> 
> Are you saying that those file should stay in `data/src`? They’ve moved to `data/covering_grammar/lib` from what I can see.

No, just that _pace_ the description they're still in-repo. I trust that y'all will place them somewhere sensible.

data/covering_grammar/lib/make_test_file.py
ajmalanoski: To be honest, I'm not very familiar with the `logging` library… But it still works if you don't specify `logging.basicConfig` --- it just uses some default parameters. This is what it looks like with setting the `basicConfig`: `WARNING:root:cat != dog (line 1)`

Is there a preferred format for logging messages in WikiPron?

kylebgorman: I'd just use the style used elsewhere nearby: https://github.com/kylebgorman/wikipron/blob/master/data/scrape/scrape.py#L174

data/covering_grammar/lib/make_test_file.py
```diff
@@ -0,0 +1,42 @@
...
+                p_word, p_pron = p_line.split("\t")
+                # Make sure that gold data and predictions have the
+                # same words.
+                if not g_word == p_word:
+                    logging.warning(
+                        "%s != %s (line %d)", g_word, p_word, lineno
+                    )
+                    continue
+                # Note that we use `strip` to remove the newline.
+                g_pron = g_pron.strip()
```
kylebgorman, line 27: If you're removing a trailing newline, use `.rstrip`. `.strip` is `.lstrip` plus `.rstrip` and we don't need `.lstrip` here I don't think.

Better yet, do it before splitting, e.g., `g_word, g_pron = g_line.rstrip().split("\t")`

data/covering_grammar/lib/make_test_file.py
```diff
@@ -0,0 +1,42 @@
...
+                p_pron = p_pron.strip()
+                line = f"{g_word}\t{g_pron}\t{p_pron}"
```
kylebgorman, line 29: Consider combining this line and the next? The variable `line` isn't reused and it's not very complicated.

kylebgorman (commented):
Few minor comments below, looking pretty good.

data/covering_grammar/lib/make_test_file.py
ajmalanoski: Sounds good. I think it would make sense drop the date/time from the config though — it seems a bit overkill for this script

kylebgorman:
Agreed, that's only really useful for long-running scripts and this is not
such a script. (There may be a model of that elsewhere.)

kylebgorman (approved):
LGTM ++

Issue #329, opened 2021-01-23
Nested lists for subdialect info


jhdeov:
I noticed this problem for [Armenian](https://en.wiktionary.org/wiki/%D5%A5%D6%80%D5%AF%D6%80%D5%B8%D6%80%D5%A4) and a colleague told me it's also found in [Portuguese](https://en.wiktionary.org/wiki/afetar). For some languages, the pronunciation entry can use a nested list, such that

* The first level contains the main dialect name
* The second level contains a subdialect name

For example, Portuguese <[afetar](https://en.wiktionary.org/wiki/afetar)> has a level-1 entry for (standard) Brazilian Portuguese. But this entry has 3 level-2 entries for different regions of Brazil. As of now, WikiPron scraps all 4 [pronunciations](https://github.com/kylebgorman/wikipron/blob/master/data/tsv/por_bz_phonemic_filtered.tsv) as part of "Brazilian Portuguese". But that obfuscates the fact that the 4 entries correspond to separate subdialects.

It would be nice if the script could 'fix' this somehow. Maybe you can add an extra column to the scraped content, such that the new column would keep the name of its line-entry's name. For example, for <afetar>, maybe you could return something like

afetar  |	a f e t a ɹ   | Brazil 
afetar  |	a f e t a ɻ   | Paulista
afetar  |	a f e t a ʁ  | South Brazil
afetar  |	a f e t a χ  | Carioca

kylebgorman:
I'd rather just have them in separate files and if people want to combine
them they're welcome to. I can't think how I'd use the three-column file
for actual processing without just filtering it into separate dialects, so
we should do that from the jump.

That said the pt issue you point out is important. We should fix that
upstream (probably).

kylebgorman:
@jacksonllee would love your thoughts on this. Keeps coming up...

jacksonllee:
> I'd rather just have them in separate files and if people want to combine
them they're welcome to. I can't think how I'd use the three-column file
for actual processing without just filtering it into separate dialects, so
we should do that from the jump.

I share the same intuition. In general, augmenting the scraped data beyond the current two-column format would be what we should avoid within WikiPron. In this particular, I agree that just having separate files for individual subdialectal varieties is the way to go.

[...]

PR #397, opened 2021-03-27
Experimental Min Nan extraction function

lfashby:

This adds an (incomplete) Min Nan extraction function in the hopes that it will nudge us toward developing solutions to #259 and #329. This extraction function currently only targets entries from the Hokkien 'dialect' (just because it seemed the most prevalent) and a user can then specify 'subdialects' of Hokkien with `--dialect`. I've added some data with (sub)dialect set as Xiamen.

To improve/expand the coverage of this extraction function we need to settle on a solution to #329. One solution might be to add a `--subdialect` option and have `--dialect` be used for Hokkein/Teochew and `--subdialect` for the nested dialects like Xiamen/Taipei (so for Portuguese we could set `--dialect` as Brazil and `--subdialect` as Paulista/South Brazil). This would be easy to implement but would probably be confusing to users. Also it would require users (or people running the big scrape) to scrape the same language a whole bunch of separate times if they were interested in getting data from all the dialects and subdialects of that language in separate TSVs.

An alternative solution would be to try and revamp our dialects system somewhat like so: If a user runs something like `wikipron nan --dialects` we run through the language once and write as many TSVs as there are dialects/subdialects in that language (easier said than done). If a user runs `wikipron nan` we run through the language once and they get one TSV containing all the entries from all the dialects/subdialects. I think there are different ways of doing this but the coolest would be to 'discover' the dialects/subdialects as we scrape a language and automatically write them to different TSVs.


kylebgorman (approved):
This looks good to me.

I am fine with the `--subdialect` proposal but agree that `--dialects` is a superior solution. It should also reduce our need to micromanage this all, right?

data/scrape/lib/languages.json
```diff
@@ -1263,7 +1263,17 @@
...
+        "casefold": true,
+        "skip_spaces_pron": false,
+        "dialect": {
+            "xi": "Xiamen"
+        },
+        "script": {
+            "zyyy": "Common",
+            "latn": "Latin",
+            "hira": "Hiragana",
+            "hani": "Han"
```
jacksonllee, line 1275: I need a quick reminder -- where do these come from again?

jacksonllee (approved):
LGTM. Kudos to you, Lucas, for working on these challenges! Understood that this Min Nan scrape was arbitrarily set for Hokkien for now, and is subject to change in future PRs based on how we want to handle subdialects.

The `--dialects` proposal does sound great (might need another flag name -- too easily confused with the existing `--dialect`), but I see implementation may be tricky (which Lucas has already alluded to). So far we've seen the Chinese-styled and Brazilian Portuguese-styled subdialect formatting on Wiktionary. Are there other flavors we haven't come across yet?

lfashby:
> It should also reduce our need to micromanage this all, right?

Certainly, though perhaps by introducing something even more demanding of micromanagement! 

> So far we've seen the Chinese-styled and Brazilian Portuguese-styled subdialect formatting on Wiktionary. Are there other flavors we haven't come across yet?

Not that I'm aware of, though it'd definitely be best to go on a bit of a hunt for them before trying out this `dialects` approach.

data/scrape/lib/languages.json
lfashby: From our new [`languages_update.py`](https://github.com/kylebgorman/wikipron/blob/master/data/scrape/lib/languages_update.py) postprocessing step.

PR #399, opened 2021-04-12
Uses context manager for file closing.

kylebgorman:
Also:

* tightens up docs
* makes mismatch a fatal error
* removes unnecessary date formatting (we don't need this because
logging isn't logging the time)


ajmalanoski:
Your edits make sense to me. LGTM (with the caveat that I'm not particularly familiar with `contextlib`).

PR #404, opened 2021-04-19
Updated the script with default token type and implementing Pynini rewrite feature

Othergreengrasses:
Updated the script with default token type and Pynini rewrite feature. Also corrected the error in the README file.

### DIFF
--- a/data/covering_grammar/lib/README.md
+++ b/data/covering_grammar/lib/README.md
@@ -5,7 +5,7 @@ This tool performs a fine-grained error analysis of a G2P model. It prints a
 
 The performance matrix is a 2x2 table where the dimensions are:
 
--   whether the hypothesized pronunciation matches the corpus prediction, and
+-   whether the hypothesized pronunciation matches the corpus prediction and
 -   whether the hypothesized pronunciation adheres to the spelling rules of the
     language and script, according to a user-provided covering grammar.
 
--- a/data/covering_grammar/lib/README.md
+++ b/data/covering_grammar/lib/README.md
@@ -27,7 +27,7 @@ Two input files are required:
     zero or more graphemes, and the right contains zero or more phones it can
     correspond to.
 2.  Test output: a three-column TSV file in which the columns are the graphemic
-    form, the hypothesized pronunciation, and the gold pronunciation.
+    form, the gold pronunciation, and the hypothesized pronunciation.
 
 ## Example workflow
 
--- a/data/covering_grammar/lib/error_analysis.py
+++ b/data/covering_grammar/lib/error_analysis.py
@@ -21,49 +21,56 @@ import argparse
 
 import prettytable  # type: ignore
 import pynini
+from pynini.lib import rewrite
 
 
 def main(args: argparse.Namespace) -> None:
-    cg_fst = pynini.string_file(args.cg_path).closure().optimize()
-    rulematch_predmatch = 0
-    rulematch_pred_notmatch = 0
-    not_rulematch_predmatch = 0
-    not_rulematch_pred_notmatch = 0
-    total_records = 0
-    with open(args.test_path, "r") as source:
-        for line in source:
-            total_records += 1
-            parts = line.split("\t")
-            lg = parts[0].rstrip()
-            act = parts[1].replace(" ", "").replace(".", "").rstrip()
-            pred_pron = parts[2].replace(" ", "").replace(".", "").rstrip()
-            # TODO: use Pynini's rewrite module here.
-            lattice = lg @ cg_fst @ pred_pron
-            lattice.project("output")
-            if lattice.start() == pynini.NO_STATE_ID:
-                if act == pred_pron:
-                    not_rulematch_predmatch += 1
+    with pynini.default_token_type("utf8"):
+        cg_fst = pynini.string_file(args.cg_path).closure().optimize()
+        rulematch_predmatch = 0
+        rulematch_pred_notmatch = 0
+        not_rulematch_predmatch = 0
+        not_rulematch_pred_notmatch = 0
+        total_records = 0
+        with open(args.test_path, "r") as source:
+            for line in source:
+                total_records += 1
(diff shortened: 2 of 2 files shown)


data/covering_grammar/lib/error_analysis.py
```diff
@@ -19,51 +19,58 @@
...
+        not_rulematch_predmatch = 0
+        not_rulematch_pred_notmatch = 0
+        total_records = 0
+        with open(args.test_path, "r") as source:
+            for line in source:
+                total_records += 1
+                (ortho_str, hypo_p_str, gold_p_str) = line.rstrip().split(
+                    "\t", 2
+                )
+                ortho = pynini.accep(ortho_str)
```
kylebgorman, line 41: I don't think you need to explicitly convert any of these to an FST since the code is running within the UTF-8 context manager. `rewrite.matches` ought to do it for you. (If I'm wrong, though I'd like to know, maybe that's something I could fix in the future. Then you can do something like:

```python
ortho, hypo_p, gold_p = line.rstrip().split("\t", 2)
if rewrite.matches(ortho, hypo_p, cg_fst):
   ...
```

Apologies in advance if I'm wrong.

I would also recommend comparing the strings (e.g., on lines 45 and 50) rather than comparing the FSTs. I'm sure comparing two FSTs is much slower than comparing two strings.

data/covering_grammar/lib/error_analysis.py
```diff
@@ -19,51 +19,58 @@
...
-    )
-    print_table.add_row(
-        ["Pron non-match", f"{rule_m_pred_nm:.2f}", f"{rule_nm_pred_nm:.2f}"]
-    )
-    print(print_table)
+                    if gold_p == hypo_p:
+                        not_rulematch_predmatch += 1
+                    else:
+                        not_rulematch_pred_notmatch += 1
+
```
kylebgorman, line 54: Please remove blank line in function. It makes it hard to tell where the function starts and ends IMO.

data/covering_grammar/lib/README.md
```diff
@@ -27,7 +27,7 @@ Two input files are required:
     zero or more graphemes, and the right contains zero or more phones it can
     correspond to.
 2.  Test output: a three-column TSV file in which the columns are the graphemic
-    form, the hypothesized pronunciation, and the gold pronunciation.
+    form, the gold pronunciation, and the hypothesized pronunciation.
```
kylebgorman, line 30: Thanks for catching this. (no action needed)

kylebgorman (commented):
One minor issue, but this looks great overall.

data/covering_grammar/lib/error_analysis.py
Othergreengrasses: Fixed.

data/covering_grammar/lib/error_analysis.py
Othergreengrasses: Fixed.

data/covering_grammar/lib/error_analysis.py
```diff
@@ -19,51 +19,54 @@
...
+                    else:
+                        not_rulematch_pred_notmatch += 1
+        # Collects percentages.
+        rule_m_pred_nm = 100 * rulematch_pred_notmatch / total_records
+        rule_m_pred_m = 100 * rulematch_predmatch / total_records
+        rule_nm_pred_m = 100 * not_rulematch_predmatch / total_records
+        rule_nm_pred_nm = 100 * not_rulematch_pred_notmatch / total_records
+        # Builds and prints the table.
+        print_table = prettytable.PrettyTable()
+        print_table.field_names = ["", "CG match", "CG not-match"]
```
kylebgorman, line 58: I think "non-match" makes more sense than "not-match", thoughts? "Not a match" is fine but "not-match" just sounds strange.

data/covering_grammar/lib/error_analysis.py
```diff
@@ -19,51 +19,54 @@
...
+                total_records += 1
+                ortho, hypo_p, gold_p = line.rstrip().split("\t", 2)
+                hypo_p = hypo_p.replace(" ", "")
+                gold_p = gold_p.replace(" ", "")
+                if rewrite.matches(ortho, hypo_p, cg_fst):
+                    if gold_p == hypo_p:
+                        rulematch_predmatch += 1
+                    else:
+                        rulematch_pred_notmatch += 1
                 else:
```
kylebgorman, line 46: Instead of doing:

```python
if A:
   if B:
     ...
   else:
     ...
else:
  if C:
    ...
  else:
    ...
```

it is (IMO) a bit simpler to just write:

```python
if A:
   if B:
     ...
   else:
     ...
elif: C:
  ...
else:
  ...
```

as was done earlier.

kylebgorman (approved):
Looks good, a few last comments.

data/covering_grammar/lib/error_analysis.py
Othergreengrasses: Addressed.

data/covering_grammar/lib/error_analysis.py
```diff
@@ -19,51 +19,53 @@
...
+        rule_nm_pred_nm = 100 * not_rulematch_pred_notmatch / total_records
+        # Builds and prints the table.
+        print_table = prettytable.PrettyTable()
+        print_table.field_names = ["", "CG match", "CG non-match"]
+        print_table.add_row(
+            ["Pron match", f"{rule_m_pred_m:.2f}", f"{rule_nm_pred_m:.2f}"]
+        )
+        print_table.add_row(
+            [
+                "Pron not-match",
```
kylebgorman, line 63: Standardize with line 57 (not-match vs. non-match).

Othergreengrasses: Fixed.

kylebgorman:
LGTM. I think you can hit the big green button ("Squash and merge") to submit; if not let me know and I'll do it for you.

Othergreengrasses:
Unfortunately, I am not seeing any merge button in this PR. Please merge this PR. :-)

PR #405, opened 2021-04-21
Minor updates for new names, adds phones postprocess

### LINKED ISSUE #360: wikipron/data/src reorganization

lfashby:

This just contains some small changes now that we are using 'broad' and 'narrow'. I don't think we really need to change anything else. Apparently our phones table was broken, so I changed the links.

I added a postprocess script to the phones side just for the sake of consistency.

Closes #360.

### DIFF
--- a/data/phones/lib/generate_phones_summary.py
+++ b/data/phones/lib/generate_phones_summary.py
@@ -39,7 +39,7 @@ def main() -> None:
     with open(LANGUAGES_PATH, "r", encoding="utf-8") as source:
         languages = json.load(source)
     readme_list = []
-    phones_summary_list = []
+    phones_summaries = []
     for file_path in os.listdir(PHONES_DIRECTORY):
         with open(
             f"{PHONES_DIRECTORY}/{file_path}", "r", encoding="utf-8"
@@ -63,17 +63,17 @@ def main() -> None:
             transcription_level,
             num_of_entries,
         ]
-        phones_summary_list.append([file_path] + row)
+        phones_summaries.append([file_path] + row)
         readme_list.append([f"[phone](phones/{file_path})"] + row)
     # Sorts by Wiktionary language name, with broad entries before narrow
     # ones.
-    phones_summary_list.sort(key=_wiki_name_and_transcription_level)
+    phones_summaries.sort(key=_wiki_name_and_transcription_level)
     readme_list.sort(key=_wiki_name_and_transcription_level)
     with open(PHONES_SUMMARY_PATH, "w", encoding="utf-8") as sink:
         tsv_writer_object = csv.writer(
             sink, delimiter="\t", lineterminator="\n"
         )
-        tsv_writer_object.writerows(phones_summary_list)
+        tsv_writer_object.writerows(phones_summaries)
     # Writes the README.
     with open(PHONES_README_PATH, "w", encoding="utf-8") as sink:
         print(
--- a/data/phones/lib/generate_phones_summary.py
+++ b/data/phones/lib/generate_phones_summary.py
@@ -5,7 +5,8 @@ import json
 import logging
 import os
 
-from typing import Any, Dict, List
+from operator import itemgetter
+from typing import Any, Dict
 
 from data.scrape.lib.codes import (
     LANGUAGES_PATH,
@@ -13,10 +14,6 @@ from data.scrape.lib.codes import (
     PHONES_README_PATH,
     PHONES_DIRECTORY,
 )
-
-
-def _wiki_name_and_transcription_level(ele: List[str]) -> str:
-    return ele[3] + ele[4]
 
 
 def _handle_wiki_name(language: Dict[str, Any], file_path: str) -> str:
@@ -67,8 +64,8 @@ def main() -> None:
         readme_list.append([f"[phone](phones/{file_path})"] + row)
     # Sorts by Wiktionary language name, with broad entries before narrow
     # ones.
-    phones_summaries.sort(key=_wiki_name_and_transcription_level)
-    readme_list.sort(key=_wiki_name_and_transcription_level)
+    phones_summaries.sort(key=itemgetter(0))
+    readme_list.sort(key=itemgetter(0))
     with open(PHONES_SUMMARY_PATH, "w", encoding="utf-8") as sink:
         tsv_writer_object = csv.writer(
             sink, delimiter="\t", lineterminator="\n"
--- a/data/phones/lib/generate_phones_summary.py
+++ b/data/phones/lib/generate_phones_summary.py
@@ -19,14 +19,10 @@ def _wiki_name_and_transcription_level(ele: List[str]) -> str:
     return ele[3] + ele[4]
 
 
-def _handle_wiki_name(
-    language: Dict[str, Any], file_path: str
-) -> str:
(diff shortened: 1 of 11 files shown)


data/scrape/tsv_summary.tsv
```diff
@@ -11,10 +11,10 @@ ale_latn_broad.tsv	ale	Aleut	Aleut	Latin		False	Broad	True	104
 grc_grek_broad.tsv	grc	Ancient Greek (to 1453)	Ancient Greek	Greek		False	Broad	True	90222
 ara_arab_broad.tsv	ara	Arabic	Arabic	Arabic		False	Broad	False	7279
 arc_hebr_broad.tsv	arc	Imperial Aramaic (700-300 BCE); Official Aramaic (700-300 BCE)	Aramaic	Hebrew		False	Broad	False	1156
+arm_armn_e_narrow.tsv	arm	Armenian	Armenian	Armenian	Eastern Armenian, standard	False	Narrow	True	14182
 arm_armn_w_narrow.tsv	arm	Armenian	Armenian	Armenian	Western Armenian, standard	False	Narrow	True	14065
 arm_armn_w_narrow_filtered.tsv	arm	Armenian	Armenian	Armenian	Western Armenian, standard	True	Narrow	True	14040
 arm_armn_e_narrow_filtered.tsv	arm	Armenian	Armenian	Armenian	Eastern Armenian, standard	True	Narrow	True	14177
```
ajmalanoski, line 17: Very minor thing, but maybe we just do finer sorting than just Wiktionary name and transcription level, since it gives  us a random sorting of the four Armenian files (all of which have the same Wiktionary name and transcription level) and of other similar cases.

ajmalanoski:
I had one minor suggestion, but otherwise LGTM. Thanks for catching the issue with the links in the HOWTO.

data/phones/lib/generate_phones_summary.py
```diff
@@ -68,34 +52,28 @@ def main() -> None:
...
         row = [
             iso639_code,
             languages[iso639_code]["iso639_name"],
             wiki_name,
             transcription_level,
             num_of_entries,
         ]
-        languages_summary_list.append([file_path] + row)
-        readme_list.append([f"[phone]({file_path})"] + row)
+        phones_summary_list.append([file_path] + row)
```
kylebgorman, line 66: drive-by, but `phones_summaries` is probably a better variable name than `phones_summary_list`.

data/phones/postprocess
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+# Runs summary generation script.
+
+set -eou pipefail
+
+main() {
```
kylebgorman, line 6: If this is just one line I think you don't have to use a `main`. I just use it when there are multiple steps.

data/scrape/tsv_summary.tsv
kylebgorman: +1

lfashby: I changed it to sort on wiki name, then whether or not they're filtered (just so that they are grouped, not really necessary but maybe looks nice), and on the path -- which has to be unique and contains transcription level anyway.

data/scrape/lib/generate_tsv_summary.py
```diff
@@ -15,8 +15,8 @@
 )
 
 
-def _wiki_name_and_transcription_level(ele: List[str]) -> str:
-    return ele[3] + ele[7]
+def wiki_name_filtered_and_path(ele: List[str]) -> str:
```
kylebgorman, line 18: Would it not suffice to just sort by the path? That'd make this function unnecessary, could do `operator.itemgetter(0)` instead, which is way faster too.

lfashby: Perhaps, but then we'd be sorting by ISO code and get things like Irish ending up next to German.

kylebgorman:
FWIW that doesn't bother me.

ajmalanoski:
I think we can close it. The original goal of the issue (i.e.,`wikipron/data/src` reorganization) has been met, IMO — `data/src` has been thoroughly reorganized (and removed).

There is still one small outstanding issue related to the project of reorganization, namely the suggested renaming/reorganization of `data/covering_grammar` (@Othergreengrasses), but I don't think it directly relates to the discussion in #360

kylebgorman:
Okay. Add it to the PR description then, Lucas, if you could.

K

lfashby:
I've hopefully sorted out our sorting problem. I can go through tomorrow and add some content to our temporary READMEs. I'll add the relevant closing statement then.

data/phones/lib/generate_phones_summary.py
```diff
@@ -68,34 +49,28 @@ def main() -> None:
...
             languages[iso639_code]["iso639_name"],
             wiki_name,
             transcription_level,
             num_of_entries,
         ]
-        languages_summary_list.append([file_path] + row)
-        readme_list.append([f"[phone]({file_path})"] + row)
+        phones_summaries.append([file_path] + row)
+        readme_list.append([f"[phone](phones/{file_path})"] + row)
     # Sorts by Wiktionary language name, with broad entries before narrow
```
kylebgorman, line 65: Is this comment still accurate?

data/phones/lib/generate_phones_summary.py
```diff
@@ -5,7 +5,8 @@
 import logging
 import os
 
-from typing import Any, Dict, List
+from operator import itemgetter
```
kylebgorman, line 8: Sort of a departure from the normal style here.

kylebgorman (approved):
LGTM, a few comments.

data/phones/lib/generate_phones_summary.py
lfashby: Good point!

lfashby:
I didn't update the README in the covering grammar section as that directory is subject to change shortly.

kylebgorman (approved):
LGTM.

kylebgorman:
Did you intend for this to be to a non-master branch? I just noticed that it is. Can I move it to master?

lfashby:
Yeah you can move it to master, it was just building on top of Aidan's changes but I think we're done renaming stuff now.

PR #420, opened 2021-05-10
Added writing an error log file in error_analysis.py

yeonju123:

I added some lines to create `logs` directory and an error log file.
I thought this will make an error analysis easier :) As for the name of the log file, I used a current timestamp. 
If you find this is useful and have a better suggestion for the logfile name, feel free to let me know :)
(Initially, I wanted to put a language name as a part of a error log file path but didn't want to create a parameter only for the name of the log file.)


data/covering_grammar/lib/error_analysis.py
```diff
@@ -18,12 +18,30 @@
...
 
+import errno
 import prettytable  # type: ignore
 import pynini
 from pynini.lib import rewrite
 
 
+def create_new_log_dir() -> str:
+    error_log_dir = os.path.join(os.getcwd(), "logs")
+    try:
```
kylebgorman, line 33: I think you can just do `exist_ok=True` as an option to `os.makedirs`. I can't think of any reason that would fail, so you can then get rid of the try/except and `errno`.

data/covering_grammar/lib/error_analysis.py
```diff
@@ -18,12 +18,30 @@
...
+import os
 
+import errno
 import prettytable  # type: ignore
 import pynini
 from pynini.lib import rewrite
 
 
+def create_new_log_dir() -> str:
+    error_log_dir = os.path.join(os.getcwd(), "logs")
```
kylebgorman, line 32: I would just call this `log`.

data/covering_grammar/lib/error_analysis.py
```diff
@@ -39,21 +57,34 @@ def main(args: argparse.Namespace) -> None:
         not_rulematch_predmatch = 0
         not_rulematch_pred_notmatch = 0
         total_records = 0
+        error_log_dir = (
```
kylebgorman, line 60: The comment seems otiose given what the function is called. Also it's causing a weird line wrap. I'd just cut it and reflow with `black -l79` again.

data/covering_grammar/lib/error_analysis.py
```diff
@@ -39,21 +57,34 @@ def main(args: argparse.Namespace) -> None:
...
-                hypo_p = hypo_p.replace(" ", "")
-                gold_p = gold_p.replace(" ", "")
-                if match_pronunciation_rule(ortho, hypo_p, cg_fst):
-                    if gold_p == hypo_p:
-                        rulematch_predmatch += 1
+            with open(
+                os.path.join(error_log_dir, today_timestamp + "_error.tsv"),
+                "w",
+                encoding="utf8",
+            ) as f:
```
kylebgorman, line 69: `f` is not a very informative variable name.

data/covering_grammar/lib/error_analysis.py
```diff
@@ -39,21 +57,34 @@ def main(args: argparse.Namespace) -> None:
...
+                writer = csv.writer(f)
```
kylebgorman, line 70: This might call for `DictWriter` instead, since you're writing a CSV with a header.

Also, why not make it a TSV? We use TSVs everywhere else in this project.

kylebgorman:
I think current directory + timestamp + `.log` is a good name. If you want to put them in a subdirectory `log` that's fine too.

kylebgorman (approved):
LGTM, thanks Yeonju!

PR #442, opened 2021-09-13
Adds test for "casefold" values in languages.json

lfashby:

A few `null` casefold values have snuck into `languages.json`. This PR updates those values and adds a test to make sure we notice when this happens in the future.

A `null` casefold value won't crash the big scrape, but it might lead to undesirable behavior. There are some potential tests we could add to the new `test_languages.py` to make sure that there is nothing in `languages.json` that will crash the big scrape (which I believe there currently is) but I'll have to think about how to do it since it might involve reorganizing `codes.py` again. If that sound interesting I can file an issue.

### DIFF
--- a/tests/test_data/test_languages.py
+++ b/tests/test_data/test_languages.py
@@ -18,8 +18,8 @@ def test_casefold_value():
         if languages[language]["casefold"] is None:
             missing_languages.add(languages[language]["wiktionary_name"])
 
-    assert len(missing_languages) == 0, (
+    assert not missing_languages, (
         "The following languages do not have a 'casefold' value "
         "in data/scrape/lib/languages.json:"
-        f"{[lang for lang in missing_languages]}"
+        f"{missing_languages}"
     )
--- a/tests/test_data/test_languages.py
+++ b/tests/test_data/test_languages.py
@@ -7,7 +7,7 @@ _REPO_DIR = os.path.dirname(
 _LANGUAGES = os.path.join(_REPO_DIR, "data/scrape/lib/languages.json")
 
 
-def test_phones_data_matches_summary():
+def test_casefold_value():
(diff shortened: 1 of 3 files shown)


kylebgorman:
> A `null` casefold value won't crash the big scrape, but it might lead to undesirable behavior. There are some potential tests we could add to the new `test_languages.py` to make sure that there is nothing in `languages.json` that will crash the big scrape (which I believe there currently is) but I'll have to think about how to do it since it might involve reorganizing `codes.py` again. If that sound interesting I can file an issue.

@lfashby do file an issue so we have notes about this---would like to know what you're thinking too.

tests/test_data/test_languages.py
```diff
@@ -0,0 +1,25 @@
...
+    has a value for 'casefold' key.
+    """
+    missing_languages = set()
+    with open(_LANGUAGES, "r") as source:
+        languages = json.load(source)
+    for language in languages:
+        if languages[language]["casefold"] is None:
+            missing_languages.add(languages[language]["wiktionary_name"])
+
+    assert len(missing_languages) == 0, (
```
kylebgorman, line 21: `assert not missing_languages`?

tests/test_data/test_languages.py
```diff
@@ -0,0 +1,25 @@
...
+        "The following languages do not have a 'casefold' value "
+        "in data/scrape/lib/languages.json:"
+        f"{[lang for lang in missing_languages]}"
```
kylebgorman, line 24: What's the point casting this to a list?

kylebgorman (commented):
Nice catch. Two nits.

kylebgorman (approved):
LGTM, thanks. I wonder if this was the issue I was stumbling on.

PR #456, opened 2021-12-18
Fix for italian broad phonemes

Frank995:
I noted an incoherence in the italian transcriptions. The broad version of the transcriptions correctly contains the phonemic representations of sounds based on the sources cited in _ita_broad.phones_. Such broad transcription correctly ignore allophones of some phonemes but **[u̯]** and **[i̯]**.

The point is, why were these phones included? They are not phonemes (differently from **/j/=[j]** and **/w/=[w]**), because no italian would ever recognize the difference between **[i] - [i̯]** and **[u]=[u̯]** unless he studied phonetic (and hardly recognize it even if he did). These should only be included in the narrow transcription.

Another point in support of this. If you were to include **[i̯]** and **[u̯]** then you would have to include also all the possible allophones of **/n/** (**[n̺], [n̪], [ŋ], [ɱ]**) which are not in the broad transcription but there are in the narrow one.

If you have any questions I will gladly answer them.

### DIFF
--- a/data/phones/phones/ita_broad.phones
+++ b/data/phones/phones/ita_broad.phones
@@ -6,35 +6,32 @@
 #   Italian, as compared with the varieties spoken in Florence. Journal of the
 #   International Phonetic Association, 35(2): 131-151.
 a
-o
-r
-e
-i
-t
-n
-l
-k
-s
-d
-m
-p
-u
-j
-ɛ
 b
-f
-v
-ɔ
-ɡ
+k
 t͡ʃ
+d
+e
+ɛ
+f
+ɡ
 d͡ʒ
+i
+j
+l
+m
+n
+o
+ɔ
+p
+r
+s
 z
-t͡s
+t
+u
 w
-ʃ
-u̯  # Allophone of /u/ in diphthongs.
+v
+t͡s
 d͡z
+ʃ
 ɲ
 ʎ
-i̯  # Allophone of /i/ in diphthongs.
-
--- a/data/scrape/tsv/ita_latn_broad.tsv
+++ b/data/scrape/tsv/ita_latn_broad.tsv
@@ -243,7 +243,7 @@ abusare	a b u z a r e
 abuso	a b u z o
 acacia	a k a t͡ʃ a
 acagiù	a k a d͡ʒ u
-acaico	a k a i̯ k o
+acaico	a k a i k o
 acalefa	a k a l ɛ f a
 acalefe	a k a l ɛ f e
 acalifa	a k a l i f a
@@ -579,7 +579,7 @@ ai	a i
 aiere	a j e r e
 ailuride	a i l u r i d e
 ailuridi	a i l u r i d i
-ailuro	a i̯ l u r o
+ailuro	a i l u r o
 ailurofobia	a i l u r o f o b i a
 ain	a j n
 aire	a i r e
@@ -645,7 +645,7 @@ alfa	a l f a
 alfabeti	a l f a b ɛ t i
 alfabetico	a l f a b ɛ t i k o
 alfabeto	a l f a b ɛ t o
-alfeide	a l f ɛ i̯ d e
+alfeide	a l f ɛ i d e
 alfiere	a l f j ɛ r e
 alga	a l ɡ a
 algarello	a l ɡ a r ɛ l l o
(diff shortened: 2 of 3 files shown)


kylebgorman:
We include these in the phonelists because the upstream source does. If we exclude them the "filtered" file won't include any words which are so transcribed. If it can be argued that only a small portion of the upstream source data contains these, such that we can treat them as an accident rather than a deliberate decision by the Wiktionarians, then we should maybe change the filtration here.

Otherwise, it seems to me like this ought to be fixed upstream (if at all); it's a case where a subphonemic detail is in the broad transcription, something that is not uncommon on Wiktionary (check out the treatment of [ŋ] in English---it's an allophone of /n/ in my analysis---or German [ç, x], which are clearly allophones of a single voiceless dorsal fricative).

Frank995:
I've got your point. I'm only talking from a "linguistics" point of view, I haven't looked yet at your code so thouroughly to know what's behind the scenes. Wouldn't be possible to add "a priori" rule to format the output?

Let's take the **[u̯]** example which should be more common than **[i̯]** in italian. Word: _augurio_ (wish), line 1506 of ita_broad.phones (there is _auguri_ which is the plural).
Wiktionary pronounciation: /au'gurjo/ (https://it.wiktionary.org/wiki/augurio)
Italian pronountiation dictionary:  /au'gurjo/ (http://dipionline.it/dizionario/ricerca?lemma=augurio)
VOCABOLaudio:  /auˈguːrjo/ (https://it.vocabolaudio.com/it/augurio)

We could argue about the length of the vowels, which is present in italian but it is not important in terms of words distinction; but it's just a matter of flow of the speech that comes natural to a native and therefore it's not always present in phonemic transcriptions.
But in none of the dictionaries I ever found those two allophones. Actually, I had never even realised we had them, even as an italian who studies phonetics :P

kylebgorman:
We don't have the feature of rules to fix transcriptions yet. I have been tempted to add something like that though, and this would be a good candidate rule.

I wonder if we can work upstream with the Wiktionarians though. I will ask someone.

Latin transcriptions have a similar problem, I believe.

Frank995:
Well for latin, the current Allen's vowels system (proposed here) is facing more and more counter arguments (paper https://drive.google.com/file/d/1xA73h_R647YZf7fp8saeRAP68x6qHa2U/view, and online discussion https://www.reddit.com/r/latin/comments/95yxez/vowel_pronunciation_beyond_allens_vox_latina/).

kylebgorman:
FWIW I was just referring to how the Wiktionary Latin transcriptions indicate the glide portion of diphthongs: /kau̯.sa/ etc. I would ust write that /kaw.sa/ myself.

Frank995:
In the broad one I would actually use /kausa/ whereas in narrow one [u̯]. In terms of perception I suppose [u̯] is more similar to [u] than [w], like in italian or sardinian. 
I'd love to know wiktionarians' point of view about this.

Frank995:
I'm having a more thourough look and definetely something should be fixed upstream. There are traces of foreigners phonemes which we do not pronounce:

- Line 5896: feuilleton	 f ø j t o n
- Line 2975: cassoeula	k a s s œ l a
- Line 13302: surf	s ə ɾ f
- And other errors here and there (and probably non-existing words)

It's clear that these are oversights and borrows from other languages but maybe somehitng should be done about it. I could work it out abit to remove as many as possible but I not sure of where to do it.

kylebgorman:
> In the broad one I would actually use /kausa/

Are you saying that the off-glide in _au_ is an allophone of /u/? That might be so but you should write out all the rules that handle allophony of /u/ before you commit to that analysis. It's not trivially correct and most theorists would just say that /w/ is an phoneme in its own right.

> whereas in narrow one [u̯]. 

Are you saying that the off-glide in _au_ is not the same phonetically as the onset glide as in, e.g., _vir_? Once again that might be so but I know of no evidence for it. Stems like _caueo_ with its perfect _cautus_ show us that coda (off-glide) and onset glides are the same phonemic element.

kylebgorman:
> * Line 5896: feuilleton	 f ø j t o n
> * Line 2975: cassoeula	k a s s œ l a
> * Line 13302: surf	s ə ɾ f

It is for precisely this reason we have phonelists. None of these three is present in our [filtered broad Italian file](https://github.com/CUNY-CL/wikipron/blob/master/data/scrape/tsv/ita_latn_broad_filtered.tsv). 

> It's clear that these are oversights and borrows from other languages but maybe something should be done about it. I could work it out abit to remove as many as possible but I not sure of where to do it.

I would edit the pronunciations if in your judgment they're not pronounced faithfully in Italian. Or if in your judgment they're not "words of Italian", you can remove them. But that's between you and the Wiktionarians. We simply mirror what's present on Wiktionary and try not to wade too deeply into this.

PR #460, opened 2022-01-26
Update `data/scrape/scrape.py`

jacksonllee:
This PR updates `data/scrape/scrape.py` to streamline the scraping workflow.

### DIFF
--- a/data/scrape/lib/README.md
+++ b/data/scrape/lib/README.md
@@ -53,11 +53,10 @@ The following steps can be used to run the big scrape procedure for a subset:
     designation and argument string must delineate with comma, semicolon, or
     space. E.g. To target only Lithuanian and Spanish:
     `./scrape.py --restriction='lit; spa'`.
-    To target all languages but several specified ones, apply the `--exclude` flag,
+    To target all but one or more specific languages, apply the `--exclude` flag,
     e.g., to exclude Lithuanian and Spanish:
     `./scrape.py --exclude='lit; spa'`.
-    Only one, but not both, of `--restriction` and `--exclude` can be concurrently
-    applied.
+    Only one, but not both, of `--restriction` and `--exclude` can be applied.
 2.  If `cut_off_date` in [`scrape.py`](../scrape.py) was set using
     `datetime.date.today().isoformat()` and it is important that all the data
     you scrape is from before the same date, then manually set `cut_off_date` in
--- a/data/scrape/lib/README.md
+++ b/data/scrape/lib/README.md
@@ -52,7 +52,12 @@ The following steps can be used to run the big scrape procedure for a subset:
     line arguments for desired languages. Note: languages must be in their ISO
     designation and argument string must delineate with comma, semicolon, or
     space. E.g. To target only Lithuanian and Spanish:
-    `./scrape.py --restriction='lit; spa'`
+    `./scrape.py --restriction='lit; spa'`.
+    To target all languages but several specified ones, apply the `--exclude` flag,
+    e.g., to exclude Lithuanian and Spanish:
+    `./scrape.py --exclude='lit; spa'`.
+    Only one, but not both, of `--restriction` and `--exclude` can be concurrently
+    applied.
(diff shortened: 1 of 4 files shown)


data/scrape/scrape.py
```diff
@@ -116,23 +116,38 @@ def _build_scraping_config(
...
-        keys = re.split(r"[;,\s]+\s*", args.restriction[0].strip(";, "))
-        if not keys[0]:
+        restriction_set = frozenset(
+            re.split(r"[;,\s]+\s*", args.restriction.strip(";, "))
+        )
+        if len(restriction_set) == 1 and not list(restriction_set)[0]:
             # Checks for empty entry.
             logging.fatal("Restriction flag raised but no language provided")
             exit(1)
-        rset = frozenset(args.restriction)
```
jacksonllee, line 128: This line was the bug that I was referring to offline. It should've been `rset = frozenset(keys)` to honor what the documentation suggests (i.e., the readme example with `--restriction='lit; spa'`).

As I continued to add the `--exclude` flag below, I refactored the code around here as well, which might have obscured what would otherwise be a one-line bug fix...

data/scrape/scrape.py
```diff
@@ -116,23 +116,38 @@ def _build_scraping_config(
...
-        if eset:
-            for key in eset:
+        if not restriction_set.issubset(codes):
+            for key in restriction_set - codes:
                 logging.fatal("%r is not a valid ISO code", key)
             exit(1)
-        codes = list(rset)
+    else:
+        restriction_set = codes
+    if args.exclude:
```
jacksonllee, line 135: I realize that this `if args.exclude` block looks very similar to the `if args.restriction` block above, but I think it's more readable as-is, rather than trying to factor things out that would add more cognitive burden to our future selves.

data/scrape/lib/README.md
```diff
@@ -52,7 +52,11 @@ The following steps can be used to run the big scrape procedure for a subset:
     line arguments for desired languages. Note: languages must be in their ISO
     designation and argument string must delineate with comma, semicolon, or
     space. E.g. To target only Lithuanian and Spanish:
-    `./scrape.py --restriction='lit; spa'`
+    `./scrape.py --restriction='lit; spa'`.
+    To target all but one or more specific languages, apply the `--exclude` flag,
+    e.g., to exclude Lithuanian and Spanish:
+    `./scrape.py --exclude='lit; spa'`.
+    Only one, but not both, of `--restriction` and `--exclude` can be applied.
```
jacksonllee, line 59: I'd prefer to keep things simple to the user / us by imposing mutual exclusion here between `--restriction` and `--exclude`, although as you'll see from my changes the underlying code technically allows their concurrent use.

requirements.txt
```diff
@@ -1,4 +1,4 @@
-black==20.8b1
+black==21.12b0
```
jacksonllee, line 1: I had to bump `black`'s version, or else the `black` check wouldn't pass for Python > 3.9.8, cf. https://stackoverflow.com/q/69912264

kylebgorman: +1

data/scrape/lib/README.md
kylebgorman: Fine with me.

My question is why not use `nargs='*'` so you can do `--restriction lit --restriction spa` or something like that? Thoughts?

data/scrape/scrape.py
kylebgorman: True. I'm not sure what I'd factor out exactly.

kylebgorman (approved):
LGTM

data/scrape/lib/README.md
jacksonllee: To get updated data for the upcoming SIGMORPHON G2P task, I was able to do something like `--restriction='<code> <code> <code> ...'` for a whole bunch of languages, rather than the otherwise unwieldy `--restriction <code>` being repeated for however many languages I needed to handle.

PR #461, opened 2022-02-04
Update data for the SIGMORPHON 2022 "known" languages

jacksonllee:
This PR made the following changes:

-   Updated scraped data in preparation for the SIGMORPHON 2022 shared task (only the "known" languages in this PR):
    `swe nno ger dut ita rum ukr bel tgl ceb ben asm per pus tha lwl`;
-   Made scripts under `data/frequencies/` and `data/morphology/` more flexible,
    especially for the purposes of preparing data for a shared task.


data/frequencies/grab_wortschatz_data.py
```diff
@@ -12,12 +13,13 @@
...
 
-WORTSCHATZ_DICT_PATH = "wortschatz_languages.json"
+_THIS_DIR = os.path.dirname(__file__)
+WORTSCHATZ_DICT_PATH = os.path.join(_THIS_DIR, "wortschatz_languages.json")
 
 
 def download(data_to_grab: Dict[str, Any]) -> Dict[str, Any]:
     to_retry = {}
-    os.mkdir("tgz")
+    os.makedirs("tgz", exist_ok=True)
```
kylebgorman, line 22: I had reason to not do it this way: I kept getting partial downloads and I couldn't figure out what was going on except via starting over. Just mentioning...

data/frequencies/grab_wortschatz_data.py
```diff
@@ -67,4 +69,10 @@ def main() -> None:
     logging.basicConfig(
         format="%(filename)s %(levelname)s: %(message)s", level="INFO"
     )
-    main()
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument(
+        "--freq-json-path",
+        default=WORTSCHATZ_DICT_PATH,
+        help="Path to the JSON file for the Wortschatz frequency download URLs",
```
kylebgorman, line 76: Nit: should be lowercase `path to...`  to match the style of autogenerated flags like for `--help` etc.

data/frequencies/merge.py
```diff
@@ -84,13 +83,34 @@ def main() -> None:
...
 if __name__ == "__main__":
     logging.basicConfig(
         format="%(filename)s %(levelname)s: %(message)s", level="INFO"
     )
-    main()
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument(
+        "--dest-dir",
+        required=True,
+        help="Destination directory where the merged data is created",
```
kylebgorman, line 101: same thing here and below: these should be lowercase to match

data/morphology/grab_unimorph_data.py
```diff
@@ -52,4 +53,10 @@ def main() -> None:
     logging.basicConfig(
         format="%(filename)s %(levelname)s: %(message)s", level="INFO"
     )
-    main()
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument(
+        "--unimorph-json-path",
+        default=UNIMORPH_DICT_PATH,
+        help="Path to the JSON file for the UniMorph download URLs",
```
kylebgorman, line 60: lowercase

data/frequencies/grab_wortschatz_data.py
kylebgorman: I might suggest instead that on running the script it should annhilate the dirs.

kylebgorman (approved):
LGTM

data/frequencies/grab_wortschatz_data.py
jacksonllee: In my workflow this time, I found it convenient to have `os.makedirs(..., exist_ok=True)`, since I used the same dirs consistently, and I just let the scripts overwrite the files inside these dirs as I iterated. Perhaps whoever prepares the shared task data next year may find another approach more convenient.

PR #471, opened 2022-11-07
remove default casefolding

jhdeov:
Removed the statement `casefold:true` from the languages.json list. I rescraped `hye` and `apw` to confirm that the languages were still scrapped, but now with the original case marking from Wiktionary. 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,7 +55,7 @@ Unreleased
 
 #### Changed
 
--   Switched to ISO 639-3 language codes. (\#468)
+-   Removed `casefold:true` from langauges.json and rescraped `hye` and `apw` (\#469)
 -   Updated scraped data in preparation for the SIGMORPHON 2022 shared task:
     `swe nno ger dut ita rum ukr bel tgl ceb ben asm per pus tha lwl`. (\#461)
 -   Made scripts under `data/frequencies/` and `data/morphology/` more flexible,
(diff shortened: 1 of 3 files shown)


data/scrape/lib/languages.json
```diff
@@ -1,55 +1,33 @@
 {
-    "aar": {
-        "iso639_name": "Afar",
-        "wiktionary_name": "Afar",
-        "wiktionary_code": "aa",
-        "casefold": true,
-        "script": {
-            "latn": "Latin"
-        }
-    },
```
jacksonllee, line 10: I'm confused -- maybe I'm missing something. Why are many entries such as this one for `aar` being deleted in this JSON file?

CHANGELOG.md
```diff
@@ -55,7 +55,7 @@ Unreleased
 
 #### Changed
 
--   Switched to ISO 639-3 language codes. (\#468)
+-   Removed `casefold:true` from langauges.json and rescraped `hye` and `apw` (\#469)
```
jacksonllee, line 58: * The new entry says `rescraped hye and apw`, but the new data isn't included in this PR?
* Why remove the existing entry `Switched to ISO 639-3 language codes. (\#468)`

jacksonllee:
If we're changing the `casefold` value for lots of languages in `languages.json`, does this mean that we'd have to re-scrape all these languages for consistency?

kylebgorman:
Yes, what @jacksonllee said: we'd want to do the full scrape. (For context, I asked @jhdeov off-thread to just pilot a few of them to see how things worked before we attempt a full scrape.)

kylebgorman:
I just implemented this in #523. Closing the PR. Thanks for the suggestion, I think you were right.

PR #480, opened 2022-11-28
MAINT update readme

jacksonllee:
Gah, I missed the incorrect Python version reminder in the README. I'm dropping it to be more future-proof...


kylebgorman (approved):
LGTM.

Yeah, that sort of non-programmatic statement tends to rot quickly. I'm fine killing it.

Issue #272, opened 2020-11-21
Text normalization in `list_phones.py`


ajmalanoski:
In the Slovenian data, some of the vowels with tone marking (e.g. /é/) are transcribed using the precomposed characters (so here, [U+00E9](https://www.compart.com/en/unicode/U+00E9) instead of the sequence [U+0065](https://www.compart.com/en/unicode/U+0065) [U+0301](https://www.compart.com/en/unicode/U+0301)). The module `ipapy` does not recognize the precomposed characters as valid IPA, so `list_phones.py` incorrectly rejects them. So, I thought it might make sense to normalize the phones as part of `list_phones.py`.

kylebgorman:
Does `ipapy` accept them in NKFD normalization? if so you could just use `unicodedata.normalize` to convert to NFKD, check against `ipapy`, and then convert back.

ajmalanoski:
I tested with several languages and only found one problem with NKFD normalization: <ç> is a valid IPA symbol, but the cedilla is not, so `ipapy.is_valid_ipa(unicodedata.normalize('NFKD', 'ç'))` returns `False`. To solve this, I think we could just check the NFKC normalization in addition to the NFKD normalization. I don't think this will create any problems here, since we're just looking at individual phones.

kylebgorman:
That could work.

1. What happens if we use NFD instead of NFKD?
2. What if we check `is_valid_ipa(nfkd_phone) or phone == "ç"`? Do you think that's simpler than trying both normalizations?

ajmalanoski:
I found two differences between the NFD and NFKD normalizations. First, if we use the NFKD normalization, then `ipapy.is_valid_ipa` will return `False` for anything with an alveolar click symbol <ǃ> in it, because that symbol gets converted to an exclamation point. Second, if we use NFKD normalization, then `ipapy.is_valid_ipa` will accept any superscript consonant or vowel symbols, including those that are not part of the standard IPA repertoire. On the other hand, the NFD normalization leads some valid IPA (e.g. /dⁿ/) to be erroneously rejected. On the whole, however, I think that the NFKD normalization leads bad superscripts to be accepted more often than NFD leads good superscripts to be rejected.

[...]

PR #497, opened 2023-07-13
Added "ː"-suffixed chars to list of valid IPA

sonofthomp:
Fixed issues outlined in [issue #272](https://github.com/CUNY-CL/wikipron/issues/272)

[list_phones.py](https://github.com/CUNY-CL/wikipron/blob/master/data/phones/lib/list_phones.py) was incorrectly rejecting IPA symbols such as "t̼ː" because ipapy doesn't recognize "ː" as a valid suffix for some IPA characters for whatever reason. I added all "ː"-suffixed IPA symbols to our supplementary list of IPA symbols not covered by ipapy to fix this.

### DIFF
--- a/data/phones/lib/list_phones.py
+++ b/data/phones/lib/list_phones.py
@@ -24,11 +24,11 @@ _other_valid_ipa = frozenset(
 )
 
 _suffixed_other_valid_ipa = frozenset(
-    phone + 'ː'
-    for phone in _other_valid_ipa
+    phone + "ː" for phone in _other_valid_ipa
 )
 
 OTHER_VALID_IPA = _other_valid_ipa | _suffixed_other_valid_ipa
+
 
 def _count_phones(filepath: str) -> Dict[str, Set[str]]:
     """Count the phones in the given TSV file.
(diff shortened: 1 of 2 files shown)


data/phones/lib/list_phones.py
```diff
@@ -17,12 +17,18 @@
 
 import ipapy
 
-OTHER_VALID_IPA = frozenset(
+OTHER_VALID_IPA = set(
```
kylebgorman, line 20: If this set is just a temporary one (i.e., isn't really used for anything but constructing `OTHER_VALID_IPA`) I probably would us `_this_kind_of_identifier` instead of `CONSTANT_IDENTIFIER`. Also, since I don't see it mutated here can it be a `frozenset`? The same thing applies to `SUFFIXED_OTHER_VALID_IPA`.

Then, one can do:

```python
OTHER_VALID_IPA = _other_valid_ipa | _suffixed_other_valid_ipa
```

which is slightly more elegant.

kylebgorman (approved):
LGTM.

kylebgorman:
Hi again, could you run `black -l79` on `list_phones.py`? That should make the remaining test pass.

sonofthomp:
Sure thing, fixed

PR #499, opened 2023-07-17
Catch iso639.language.LanguageNotFoundError

sonofthomp:
The iso639 module occasionally doesn't recognizing ISO language codes. Updated the code to catch this error and display a warning.

### DIFF
--- a/data/scrape/lib/codes.py
+++ b/data/scrape/lib/codes.py
@@ -174,14 +174,14 @@ def main() -> None:
             wiktionary_code = _scrape_wiktionary_language_code(
                 wiktionary_name.replace(" ", "_")
             )
-            
+
             try:
                 iso639_lang = iso639.Language.match(wiktionary_code)
             except iso639.language.LanguageNotFoundError:
                 logging.warning(
                     "Could not find language with code %s", wiktionary_code
                 )
-                
+
             if iso639_lang is None:
                 # No match found for the Wiktionary code.
                 unmatched_languages[wiktionary_code] = {
--- a/data/scrape/lib/codes.py
+++ b/data/scrape/lib/codes.py
@@ -174,7 +174,14 @@ def main() -> None:
             wiktionary_code = _scrape_wiktionary_language_code(
                 wiktionary_name.replace(" ", "_")
             )
-            iso639_lang = iso639.Language.match(wiktionary_code)
+            
+            try:
+                iso639_lang = iso639.Language.match(wiktionary_code)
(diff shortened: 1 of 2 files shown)


sonofthomp:
Fixing checks rn.

data/scrape/lib/codes.py
```diff
@@ -174,7 +174,14 @@ def main() -> None:
             wiktionary_code = _scrape_wiktionary_language_code(
                 wiktionary_name.replace(" ", "_")
             )
-            iso639_lang = iso639.Language.match(wiktionary_code)
+
```
kylebgorman, line 177: Check looks good. Don't you need some kind of `continue` statement to exit the loop, or a default value for `iso_639_lang`? Otherwise if you hit the `except` clause, it looks to me like the identifier `iso639_lang` will be undefined (or take the value of the previous loop given Python's scoping rules). But that's just my conjecture from reading the function...I may be wrong.

data/scrape/lib/codes.py
```diff
@@ -174,7 +174,14 @@ def main() -> None:
...
+            try:
+                iso639_lang = iso639.Language.match(wiktionary_code)
+            except iso639.language.LanguageNotFoundError:
+                logging.warning(
+                    "Could not find language with code %s", wiktionary_code
+                )
+
```
kylebgorman, line 184: This is admittedly a pet peeve of mine but I absolutely loathe blank lines within functions; they make it hard for me to see where one functions begins and another ends. Can you remove the new ones on lines 177 and 184? If you want to set off the code here from the rest of the function, a comment would be fine.

kylebgorman (commented):
Thanks! Two notes.

data/scrape/lib/codes.py
sonofthomp: This is a good catch, I'll fix it.

This actually hadn't caused an error in the loop when I tested the program, because it was taking on the value of iso639_lang from the previous iteration. Glad you noticed it

data/scrape/lib/codes.py
sonofthomp: Sure, np

data/scrape/lib/codes.py
kylebgorman: Yeah, things like that are why I hate Python's weird scope rules...

sonofthomp:
Should be good now

kylebgorman (approved):
LGTM

PR #508, opened 2023-08-25
Moves installable module into `src` subdirectory.

kylebgorman:
This prevents the installation of `data`, `test`, etc. modules.


kylebgorman:
This breaks some tests, any suggestions @jacksonllee?

jacksonllee:
The black, flake8, and mypy checks failed. Their commands in `.circleci/config.yml` need to be updated as well due to the directory change, e.g., `flake8 wikipron tests` -> `flake8 src tests` at [this line](https://github.com/CUNY-CL/wikipron/blob/714bd6b34a25c11c544f290bf486dbf0eb141dae/.circleci/config.yml#L77).

kylebgorman:
> The black, flake8, and mypy checks failed. Their commands in `.circleci/config.yml` need to be updated as well due to the directory change, e.g., `flake8 wikipron tests` -> `flake8 src tests` at [this line](https://github.com/CUNY-CL/wikipron/blob/714bd6b34a25c11c544f290bf486dbf0eb141dae/.circleci/config.yml#L77).

Sorry to waste your time here. I meant to draw your attention to the failing `pytest` tests (which I knew about from running locally) not the failing `flake8`, etc. ones. 

The problem is that the unit tests call complicated functions from deep within `data` which now no longer works because it's not an installed module. But, making the import paths like `data.scrape.lib...` breaks our ability to do things like run `./scrape.py` inside `data/scrape`, so I'm not sure what the right course of action is here.

jacksonllee:
Locally, I pulled down your branch and tried the following relative (as opposed to absolute) imports, which appeared to do the trick for making all the tests pass (and which _shouldn't_ impact our ability to run things like `./scrape.py` under `data/scrape/`):

* `from lib.codes ...` -> `from .lib.codes ...`, [here](https://github.com/CUNY-CL/wikipron/blob/06632fa26d454604003a0b69807aad18c79f6098/data/scrape/scrape.py#L15)
* `from codes ...` -> `from .codes ...`, [here](https://github.com/CUNY-CL/wikipron/blob/06632fa26d454604003a0b69807aad18c79f6098/data/scrape/lib/languages_update.py#L20) and also [here](https://github.com/CUNY-CL/wikipron/blob/06632fa26d454604003a0b69807aad18c79f6098/data/scrape/lib/split.py#L15)

kylebgorman:
> Locally, I pulled down your branch and tried the following relative (as opposed to absolute) imports, which appeared to do the trick for making all the tests pass (and which _shouldn't_ impact our ability to run things like `./scrape.py` under `data/scrape/`):
> 
> * `from lib.codes ...` -> `from .lib.codes ...`, [here](https://github.com/CUNY-CL/wikipron/blob/06632fa26d454604003a0b69807aad18c79f6098/data/scrape/scrape.py#L15)
> * `from codes ...` -> `from .codes ...`, [here](https://github.com/CUNY-CL/wikipron/blob/06632fa26d454604003a0b69807aad18c79f6098/data/scrape/lib/languages_update.py#L20) and also [here](https://github.com/CUNY-CL/wikipron/blob/06632fa26d454604003a0b69807aad18c79f6098/data/scrape/lib/split.py#L15)

That fixes the tests but breaks `./scrape.py`:

```
ImportError: attempted relative import with no known parent package
```

I think the solution is going to have to be to move just about all the code to `data/scrape/lib` and then `data/scrape/scrape.py` will just invoke, say, `data.scrape.lib.scrape.main()` or something like that. Working on that now.

kylebgorman:
I now have both uses working. To do this I had to create `lib/scrape.py`, and for consistency, it is called from a newly created `./scrape` Bash script.

I can no longer repeat the directory/pathname hacking code in one location, because, e.g., `split.py` can't see `codes.py` in this setup. It's a bit of a pain but I don't see any good alternative.

kylebgorman:
Hi @jacksonllee, could I get your thoughts on this design?

jacksonllee (approved):
LGTM!

PR #526, opened 2024-03-12
Adding a phonefilter for Bengali (Rarhi variety) Closes #521

### LINKED ISSUE #521: [ben] phonelist
We do not have a phonelist for Bengali and since we're using it for work in other projects, it seems like we ought to.

Othergreengrasses:
Closes #521

Please merge if it looks good.


kylebgorman (commented):
The phonelist itself looks good to me.

1. I cleaned up the comments to match style elsewhere. Seems better to just give the proper citations at the top than repeatedly throughout. 
2. You need to run `./postprocess` in the `phones` directory when you add a new phonelist, and add the README and the summary TSV file to the PR. 
3. Add the Rarhi dialect [here](https://github.com/CUNY-CL/wikipron/blob/fe97228d4c3d1e79a7afcb7096da0fc24b81e2d5/data/scrape/lib/languages.json#L285). For the format, see the `dialect` keys elsewhere in the file. Then add that file to the PR. 
4. Re-scrape Bengali itself: `./scrape --restriction ben` then `./postprocess` then add the newly modified README and TSV files to the PR.

It might be wise to add the other dialects into this PR, too. Then you'll want to repeat steps 2-4.

kylebgorman:
Please add the string "Closes #521" to the description of the PR above (it's editable too) so the issue and bug are bidirectionally linked to each other.

kylebgorman:
> Please add the string "Closes #521" to the description of the PR above (it's editable too) so the issue and bug are bidirectionally linked to each other.

Please do this; otherwise LGTM and very nice. I am seeing a test issue but I think merging with HEAD will solve that (I'll test locally before I merge).

Othergreengrasses:
Added Closes #521. Thank you!

PR #529, opened 2024-04-08
Fixes excessive line wrapping.

kylebgorman:
The Min Nan `nan` smoke test is broken. It seems that Min Nan has been removed from English Wiktionary, without a trace. So, I have proposed to delete its custom selector. @jacksonllee if you have a moment, could you make sure I'm not making things worse? (Is it perhaps now recorded as Hokkien or Fujian?)


data/requirements.txt
```diff
@@ -1,7 +1,8 @@
 ipapy>=0.0.9.0
+lxml_html_clean
```
jacksonllee, line 2: Were you trying to use lxml_html_clean somewhere? It's not used in this pull request.

jacksonllee (approved):
Looks great overall. A small merge conflict for `CHANGELOG.md` needs to be resolved. I also have a question about "lxml_html_clean".

> The Min Nan nan smoke test is broken. It seems that Min Nan has been removed from English Wiktionary, without a trace. So, I have proposed to delete its custom selector. @jacksonllee if you have a moment, could you make sure I'm not making things worse? (Is it perhaps now recorded as Hokkien or Fujian?)

Everything looks right to me with respect to dropping Min Nan here. Hokkien would be a good addition, but [it currently has only 74 entries on Wiktionary](https://en.wiktionary.org/wiki/Category:Hokkien_terms_with_IPA_pronunciation). When it reaches our threshold of 100, we can pull in the data.

data/requirements.txt
kylebgorman: See note here: 

https://github.com/CUNY-CL/wikipron/pull/529/commits/21054c3be6227f8fb7184ab9194f42465f03107a

Tests fail without it, and including it is the suggested solution.

PR #523, opened 2024-03-06
Big scrape: no casefolding

kylebgorman:
This responds to #469 and #471 by removing casefolding from the big scrape. If accepted, the big scrape will also be re-run with this in place.

- [x] Updated `Unreleased` in `CHANGELOG.md` to reflect the changes in code or data.

Issue #514, opened 2024-01-31
[lat] Broken selector


kylebgorman:
As of at least #509 the custom selector for Latin has been broken.

Latin has a [custom selector](src/wikipron/extract/lat.py) because the headwords lack macrons. Now the Romans of course didn't use macrons (and they did not consistently indicate vowel length) but just about every modern-era pedagogical resource does, so this was frankly a bizarre decision by the editors, and requires us to find the macronized forms somewhere else on the page (namely in the etymology subsection), then merge these together with the pronunciations. To debug, the obvious thing to do is to mock up one or the other stream: the etymological macronized headwords, or the pronunciations, and see which one isn't matching.

It should be possible to push through with a new version of #509 and just ignore Latin, which is probably a good idea given that it's been a while since a big scrape has been merged.

kylebgorman:
Testing of Latin is (hackily) paused in #520.

PR #536, opened 2024-05-01
Big scrape after #523

kylebgorman:

#523 removes casefolding from the big scrape, which should give us substantially more entries in the languages that use casing informatively. This PR is the result of simply rerunnning the big scrape after #523 (and any other subsequent changes). 

Note that because #514 for broken selectors in Latin is outstanding, I have not updated that data.


kylebgorman:
I'm having trouble manually merging this so I will start a new PR. Sorry for the noise.

PR #542, opened 2024-05-29
Added geminate consonants to Norwegian Bokmal phonelist

### LINKED ISSUE #532: [nob] overly restrictive phonelist
In Norwegian Bokmål our phonelist excludes geminate consonants, which are robustly attested; e.g., _tː_.

fhallee:
Added long consonants Norwegian Bokmal (nob) phonelist, as they were missing.

Closes #532


kylebgorman:
The change to the phonelist itself looks good. Please follow steps 4-7 [here](https://github.com/CUNY-CL/wikipron/blob/master/data/phones/HOWTO.md#how-to-submit-a-phones-file) and the tests you see above ought to pass. 

I am assuming that segments like /h/ are excluded on the grounds they are (never?) geminates in this language.

fhallee:
> I am assuming that segments like /h/ are excluded on the grounds they are (never?) geminates in this language.

That’s right, I only considered consonants for which I found evidence, either online or in the data itself, that they can be geminates. /h/ was not one of these. Unfortunately, there wasn't a lot of detailed information online. Many sources simply mentioned the possibility of "long consonants” and their relationship to short vowels.

kylebgorman:
> That’s right, I only considered consonants for which I found evidence, either online or in the data itself, that they can be geminates. /h/ was not one of these. 

That’s fine, proceed with that.

data/scrape/lib/unscraped.json
```diff
@@ -0,0 +1,4 @@
+{
```
kylebgorman, line 1: Please run `git checkout -- unscraped.json` (in this directory, or use the full path from another directory). This doesn't need checked in.

data/scrape/lib/languages.json
```diff
@@ -460,7 +460,8 @@
             "latn": "Latin",
             "hira": "Hiragana",
             "bopo": "Bopomofo",
-            "hani": "Han"
+            "hani": "Han",
```
kylebgorman, line 463: Please run `git checkout -- languages.json` (in this directory, etc.) on this file. It doesn't need checked in or modified here, even if the procedures caused it to be. (I think that's a very minor bug for us to fix separately.)

kylebgorman:
Reverting those two files (as suggested above) and then rerunning `./postprocess` in both the `scrape` and `phones` directory ought to get the tests passing again---try it out.

kylebgorman:
I realize I mispoke. I think that `git checkout` won't undo a change you've already committed, it will just undo an uncommitted change? That is why `languages.json` and `unscraped.json` are still in this PR. I had to research how to do this. The internet says to try the newly-added (as of late 2019) verb `restore`. That is: `git restore myfile` will restore `myfile` to its state at HEAD. After committing and pushing, you'll know it worked if those files are no longer listed in the PR on GitHub.

fhallee:
Restore didn’t seem to do anything, but I was able to remove the two files from the PR by reverting the files to the commit before I made the changes using `git checkout <commit-hash> <file-name>` and then committing those changes

kylebgorman:
There is now a different, later error that the tests are failing on, but it's one completely unrelated to what you did so I give you a dispensation to submit; I'll resolve that issue separately. Let me know if you're ready for me to merge.

fhallee:
Yep, ready!

kylebgorman:
Before I merge this I'm going to edit your PR description to include the string "Closes #532". This tells GitHub that the PR, once merged, should be linked with that issue, and should mark it as closed.

PR #578, opened 2026-04-09
[lat] Update Latin XPath selectors

### LINKED ISSUE #514: [lat] Broken selector

jacksonllee:
This pull request updates the Latin XPath selectors due to upstream Wiktionary HTML changes. The previously commented-out smoke test for Latin is now re-enabled, though apparently the current Latin data is only narrow transcription now.

I was going to maybe also do a scrape run just for Latin to go with the PR, but then I noticed a couple things that we should optimize before running any scraping -- another PR's coming up for these.

This PR resolves #514.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ Unreleased
 
 ### Under `src/` and elsewhere
 
+-   Updates Latin XPath selectors. (\#578)
 -   Replaces `--skip-parens`/`--no-skip-parens` with `--parens` accepting
     `skip`, `show`, or `expand`. The `expand` option generates all
     pronunciation variants from parenthesized content and is the new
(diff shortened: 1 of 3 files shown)


kylebgorman (approved):
LGTM

PR #579, opened 2026-04-09
Optimize scraping

jacksonllee:
This pull request optimizes scraping and cleans up a few related things across the codebase.

For the core scraping:

- Exponential backoff: A flat 300s sleep whenever there's a hiccup with the MediaWiki API was too inefficient. It's now replaced with exponential backoff (5s base, 300s cap, with jitter).
- Session reuse: One `requests.Session` for the entire scrape run (both API
  pagination and page fetches), instead of a new `HTMLSession` per 500-word
  batch + bare `requests.get()` for API calls
- API timeout: Added `timeout=30` to the API call (previously had none)

While I looked into the way API calls are made, I saw we were using the unmaintained requests_html wrapper package. I decided to switch to directly using requests and lxml ourselves, with the additional benefit of dropping unnecessary transitive dependencies (e.g., pyppeteer, which downloads Chromium...). Also added the stub packages types-requests and types-lxml as dev deps to make mypy happy.

And because I updated the dependencies:
- Everything under `data/` including both the scraped data and all the tools related to running the scraping has become as integral as the "core" under `src/` as part of WikiPron. So I consolidated the dependencies by removing `data/requirements.txt` and having the top-level `pyproject.toml` + `requirements.txt` manage the full set of dependencies.
- The latest numpy (v2.4.4), a transitive dependency of pandas, no longer has a wheel for Python 3.10 from PyPI. Instead of fighting with numpy/pandas version for Python 3.10, a much easier solution is to simply drop Python 3.10 entirely from our end. I don't think we'd _need_ Python 3.10 for WikiPron at this point.
- Since uv has practically taken over the Python world, I've updated docs for codebase setup.

After this PR is merged, I'll attempt a scrape run for Latin.


.circleci/config.yml
```diff
@@ -95,7 +95,7 @@ workflows:
             - twine
           matrix:
             parameters:
-              python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
+              python-version: ["3.11", "3.12", "3.13", "3.14"]
```
kylebgorman, line 98: I'm fine with this.

src/wikipron/html_utils.py
```diff
@@ -0,0 +1,50 @@
+"""Thin wrapper over requests + lxml for HTML fetching and XPath queries.
```
kylebgorman, line 1: Could I suggest you type-declare all the interfaces in this library? Shouldn't be hard.

src/wikipron/scrape.py
```diff
@@ -25,6 +26,10 @@
     ),
 }
 
+# Exponential backoff parameters for retrying after connection errors.
+_BACKOFF_BASE = 5  # seconds
+_BACKOFF_MAX = 300  # seconds
+
```
kylebgorman, line 33: Maybe give the formula in a comment here; I think it's:

$$\max(b r^2 + \textrm{runif}(0, 1), m)$$

where b is base, r is the number of retries, and m is the max.

pyproject.toml
```diff
@@ -33,14 +38,26 @@ classifiers = [
...
+dev = [
+    'black',
+    'build',
+    'flake8',
+    'mypy',
+    'pytest',
+    'setuptools',
+    'twine',
+    'types-lxml',
+    'types-requests',
```
kylebgorman, line 58: add `uv` to this since you're recommending it?

kylebgorman (commented):
All seems good; I assume the proof is in the scrape pudding though. 

A few comments below, otherwise LGTM.

pyproject.toml
jacksonllee: uv is an external tool (in the same category along its competitors like pip or conda), should be installed before any of these core or dev deps for WikiPron are installed, and shouldn't be specified in a dependency list in pyproject.toml.

jacksonllee:
@kylebgorman This PR is ready for another review.

kylebgorman (approved):
LGTM

PR #581, opened 2026-04-16
[lat] Update Latin data; implement multi-config scraping

jacksonllee:
This pull request updates the Latin data with a Latin-specific scraping run. Also optimizes the scraping pipeline and fixes a couple of issues that came up during the scraping run.

* Latin: There is no longer broad transcription from the upstream data. All "broad" TSVs are removed.
* Multi-config scraping: Previously, the scraping pipeline under `data/scrape/` fetched a Wiktionary page multiple times for all combinations of broad/narrow and dialects, thereby making unnecessary HTTP requests to the MediaWiki API. Multi-config scraping has now been implemented to only fetch a given page once and parse it for multiple configurations (broad/narrow, dialects).


data/phones/phones/lat_clas_broad.phones
jacksonllee, line 1: I'm 50/50 for keeping or removing this file. It wouldn't do any harm, so I'm leaving this comment here, but let me know if you prefer to remove it instead.

src/wikipron/extract/default.py
```diff
@@ -12,7 +12,7 @@
     from wikipron.typing import Iterator, WordPronPair
 
 
-IPA_XPATH_SELECTOR = '//span[contains(@class, "IPA")]'
+IPA_XPATH_SELECTOR = './/span[contains(@class, "IPA")]'
```
jacksonllee, line 15: This change fixes a bug due to the switch from the requests_html package to requests + lxml. requests_html silently targets a specific element, so here we need this `.` to avoid parsing from the document root all the time (or else we'd pull in all pronunciations from all languages on a given Wiktionary page).

src/wikipron/scrape.py
```diff
@@ -129,3 +137,12 @@ def scrape(config: Config) -> Iterator[WordPronPair]:
             )
             time.sleep(delay)
             retries += 1
+
+
+def scrape(config: Config) -> Iterator[WordPronPair]:
```
jacksonllee, line 142: No changes to the core `scrape` function. This is just refactoring to expose some of the inner iterators for the implementation of multi-config scraping under `data/scrape/`.

data/phones/phones/lat_clas_broad.phones
kylebgorman: Seems harmless to keep, also fine to remove since it'll always be in the history. Up to you.

data/scrape/lib/scrape.py
```diff
@@ -2,16 +2,19 @@
...
 import os
 import re
+import unicodedata
 
 from collections.abc import Iterator
 from typing import Any
 
-import wikipron  # type: ignore
+import wikipron
+from wikipron.scrape import iter_page_responses
```
kylebgorman, line 17: Maybe you want to do `from wikipron import scrape` to match the general style? But maybe we have also drifted a bit from that style at this point.

kylebgorman (commented):
The retooling of the scrape seems fine to me. **One question I have about this**: is it even possible to define `languages.json` so that you have incompatible configs of the sort you check for? 

I also looked at the data a little. I may have confused you off-thread, or confused myself. I did a quick check with `lat_latn_clas_narrow.tsv`, which I expected to have many Classical, narrowly transcribed forms. I searched for orthographic words matching /^amā/, but to my surprise, I did not get any inflected forms of _amō_, _amāre_, _amāvī_, _amātum_ 'love' as I expected. Here is what I actually see:

```
$ grep -E '^amās' lat_latn_clas_narrow.tsv
amāsia  a m aː s i a
amāsie  a m aː s i ɛ
amāsiuncula     a m aː s i ʊ ŋ k ʊ ɫ a
amāsiunculus    a m aː s i ʊ ŋ k ʊ ɫ ʊ s
amāsius a m aː s i ʊ s
amāsiō  a m aː s i oː
amāsī   a m aː s iː
```

This is despite the fact that many of these inflected forms of the verb have Classical Latin narrow pronunciations on Wiktionary (see [here](https://en.wiktionary.org/wiki/amo#Conjugation)). Maybe I mispoke when I said that one wants to pull in the lemma; I was trying to differentiate between the ASCII-only Wiktionary key and the diacriticized word, but I didn't mean to imply that we really want the lemma in the morphological sense here. I think we'd love to pull in all the inflected forms of _amō_ that have Classical narrow pronunciations. Sorry for confusion if that's what's going on.

jacksonllee:
@kylebgorman This PR is ready for review again.

> The retooling of the scrape seems fine to me. One question I have about this: is it even possible to define languages.json so that you have incompatible configs of the sort you check for?

Not possible from `languages.json`, though from the lens of the `scrape_multi` function alone feeding in incompatible configs would technically be possible, hence the guardrail.

Re: getting all the inflected forms rather than just the morphological lemma -- No, you didn't confuse me. The issue was the XPath selectors that didn't target the right things. I've fixed them and done a new scrape run for Latin yesterday. If you pull down this branch and check out the new data, you should see all of _amō, amāre, amāvī, amātum_, etc.

data/scrape/lib/scrape.py
jacksonllee: We need to do `from wikipron.scrape import iter_page_responses`, because `from wikipron import scrape` would import the function (not module) `scrape` (defined [here](https://github.com/CUNY-CL/wikipron/blob/185dfcf514fe37fe4cfddddeacff1eda3afdca55/src/wikipron/scrape.py#L88), then re-exported [here](https://github.com/CUNY-CL/wikipron/blob/185dfcf514fe37fe4cfddddeacff1eda3afdca55/src/wikipron/__init__.py#L6) under the `wikipron` namespace).

kylebgorman (approved):
LGTM

PR #582, opened 2026-04-17
Sort `languages.json` and `unmatched_languages.json` by key

jacksonllee:
The [language coverage test](https://github.com/CUNY-CL/wikipron/blob/11419a13c180153c094c1a308b438bb1ff67d4f5/tests/test_wikipron/test_languagecodes.py#L17) has flagged that there are a few dozen new languages we could add to WikiPron. To facilitate adding these languages, it would be ideal to have sorted both `languages.json` and `unmatched_languages.json` by key, so that the diffs would be a lot easier to review moving forward. This PR sorts these two JSON files with no behavioral changes; the new languages will come in the next PR.

### DIFF
--- a/data/scrape/lib/codes.py
+++ b/data/scrape/lib/codes.py
@@ -202,10 +202,11 @@ def main() -> None:
                     **core_settings,
                 }
             _check_language_code_against_wiki(iso639_code, wiktionary_name)
+    dump_kwargs = {"indent": 4, "ensure_ascii": False, "sort_keys": True}
     with open(LANGUAGES_PATH, "w", encoding="utf-8") as sink:
-        json.dump(new_languages, sink, indent=4, ensure_ascii=False)
+        json.dump(new_languages, sink, **dump_kwargs)  # type: ignore[arg-type]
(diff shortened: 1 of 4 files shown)


data/scrape/lib/codes.py
```diff
@@ -202,10 +202,11 @@ def main() -> None:
                     **core_settings,
                 }
             _check_language_code_against_wiki(iso639_code, wiktionary_name)
+    dump_kwargs = {"indent": 4, "ensure_ascii": False, "sort_keys": True}
```
kylebgorman, line 205: Just one man's feeling but I don't think this is an improvement given that you have to fight the typing system. I'd just copy-paste...or you could write a helper function if you really want.

kylebgorman (approved):
LGTM. One comment but it's not important ;)

data/scrape/lib/codes.py
jacksonllee: Yeah, none of these options would make the mypy gods any happier. There's a non-zero chance that I may have to tinker with this code again when working through the new languages, so I'm merging as-is for now.

PR #585, opened 2026-05-07
Support newly available languages from upstream

jacksonllee:
This pull request handles languages flagged by the `test_language_coverage` check in the test suite --
* 29 new-to-Wikipron languages: Added support for these languages and scraped their data;
* 4 languages for which Wikipron has had data, but upstream Wiktionary has changed their names. New data was also scraped for them.

See the changelog entry for what these languages are.


data/scrape/lib/unmatched_languages.json
```diff
@@ -1,4 +1,7 @@
 {
+    "art-bsp": {
```
kylebgorman, line 2: Is this a conlang? Do we want conlangs?

kylebgorman (approved):
LGTM; one comment.

data/scrape/lib/unmatched_languages.json
jacksonllee: > Do we want conlangs?

I don't think so. This `unmatched_languages.json` file is an artifact of `data/scrape/lib/codes.py`, and isn't read by anything in the codebase. It's purely documentation/bookkeeping of "things Wiktionary has that we can't match to ISO 639-3". As the codebase stands currently, none of these "unmatched" languages can be scraped, for two reasons:

* None of them are in `data/scrape/lib/languages.json` (read by `data/scrape/lib/scrape.py`, the entry-point of the a scrape run);
* None of them are handled in the `src/wikipron/languagecodes.py::LANGUAGE_CODES` mappings (used by the core `wikipron.scrape` function to resolve a language name).

We could certainly consider including the non-conlangs in `unmatched_languages.json` (by adding them to `languages.json` and `LANGUAGE_CODES`) -- I can check them out later (currently dealing with other languages already in `data/scrape/tsv/`).

kylebgorman:
Oh right. I forgot this isn’t a list of languages to be scraped. All good.

PR #588, opened 2026-05-12
[eng, fin, por, spa] Data update

jacksonllee:
This PR updates the TSV data for the following languages:
* English (`eng`)
* Finnish (`fin`)
* Portuguese (`por`)
* Spanish (`spa`)


.github/dependabot.yml
```diff
@@ -0,0 +1,9 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "."
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 0
+    ignore:
+      - dependency-name: "*"
```
jacksonllee, line 9: Temporarily disallowing GitHub's dependabot to create PRs for `requirements.txt` due to occasional CVEs. Some thoughts:

* Ideally, this `requirements.txt` should be properly refreshed by the `uv pip compile ...` command documented at the top of the file, so that uv's dependency version resolver has a chance to make sure the entire set of dependencies and their versions are compatible with one another. The dependabot's PRs target specific dependency versions in a piecemeal fashion, which might result in a broken Python environment due to incompatible dependencies.
* The dependabot's PRs have become a little more noisy than helpful lately. Also, for our use cases, a slightly stale `requirements.txt` would pose little or no security risks. So I think it's perfectly fine to silence this bot for now.
* There may be a way to make the dependabot and a GitHub Actions workflow work together, so that a CVE would automatically trigger running the uv command to refresh `requirements.txt` for us. I haven't explored this yet, though honestly this isn't a priority.

data/scrape/lib/languages.json
jacksonllee, line 1: Sorting this `languages.json` file by key, also adding the kwarg `sort_keys=True` in `languages_update.py` below that generates this JSON file.

kylebgorman (approved):
LGTM

kylebgorman (approved):

PR #593, opened 2026-07-22
[ben, cym, eng, hye, lat, por, spa, tha, vie] Fix dialect XPath selectors

jacksonllee:
I noticed that common words like "the" (!) are missing from the English TSVs -- it's upstream Wiktionary changes (again). This PR fixes the impacted dialect XPath selectors and updates all languages that have dialect configurations in the big scrape.

For our reference, the English data broke at the #588 re-scrape (2026‑05‑12), although "the" had actually been present since the #568 scrape (2025‑04‑20) with the same (previous) dialect selector -- now fixed/updated by this PR.


data/scrape/lib/verify_scrape.py
```diff
@@ -0,0 +1,277 @@
...
+    ./lib/verify_scrape.py tsv/tha_thai_broad.tsv
+
+That compares the working-tree file (NEW) against its committed (HEAD)
+version (OLD). To compare against a different commit or an explicit
+file::
+
+    ./lib/verify_scrape.py tsv/tha_thai_broad.tsv --against v2.2.0
+    ./lib/verify_scrape.py tsv/tha_thai_broad.tsv --old /tmp/old.tsv
+
+To tell genuine Wiktionary attrition from a partial scrape, probe a
```
kylebgorman: "To detect"?

data/scrape/lib/verify_scrape.py
```diff
@@ -0,0 +1,277 @@
...
+from wikipron.scrape import HTTP_HEADERS, _PAGE_TEMPLATE
+
+
+def _nfc(text: str) -> str:
+    return unicodedata.normalize("NFC", text)
+
+
+def _read_working(path: str) -> list[str]:
+    with open(path, encoding="utf-8") as source:
+        return source.read().splitlines()
```
kylebgorman: `.readlines()` does this

data/scrape/lib/verify_scrape.py
```diff
@@ -0,0 +1,277 @@
...
+    shown = subprocess.run(
+        ["git", "-C", toplevel, "show", f"{ref}:{rel}"],
+        capture_output=True,
+        text=True,
+        check=True,
+    )
+    return shown.stdout.splitlines()
+
+
+def _pairs(lines: list[str]) -> list[tuple[str, str]]:
```
kylebgorman: consider making this a generator?

data/scrape/lib/verify_scrape.py
```diff
@@ -0,0 +1,277 @@
...
+        absent = word not in new_words
+        if absent and start is None:
+            start = i
+        elif not absent and start is not None:
+            runs.append((start, i - 1))
+            start = None
+    if start is not None:
+        runs.append((start, len(old_order) - 1))
+    sized = [(old_order[a], old_order[b], b - a + 1) for a, b in runs]
+    sized.sort(key=lambda run: -run[2])
```
kylebgorman: I would like to suggest a better idiom for this but will defer to you:

`sized.sort(key=operator.itemgetter(2), reverse=True)`

data/scrape/lib/verify_scrape.py
```diff
@@ -0,0 +1,277 @@
...
+    print(f"  common words:              {len(common)}")
+    print(f"  common words, pron changed:{len(changed)}")
+
+    runs = _dropped_runs(_ordered_unique_words(old_pairs), new_words)
+    big = [r for r in runs if r[2] >= args.min_run]
+    print(f"\ncontiguous dropped-runs (>= {args.min_run} words): {len(big)}")
+    for first, last, length in big[:10]:
+        print(f"    len={length:>4}   [{first} .. {last}]")
+
+    partial_suspected = bool(big)
```
kylebgorman, line 213: Consider an implicit coercion:

`if big:`

or the more explicit, also constant-time:

`if len(big) > 0:`

tests/test_wikipron/test_scrape.py
```diff
@@ -84,6 +87,50 @@
 ]
 
 
+# Per-language wall-clock budget for the live-API smoke tests. A language that
+# stops yielding (e.g. after a Wiktionary layout change) would otherwise walk
+# its entire category and hang CI past the 10-minute no-output limit.
+_SMOKE_TEST_TIMEOUT_SECONDS = 60
+
+
+def _collect_pairs(config, n, timeout):
```
kylebgorman: consider typing this

kylebgorman (commented):
A few comments...cool though.

data/scrape/lib/verify_scrape.py
kylebgorman: or make it a generator

data/scrape/lib/verify_scrape.py
jacksonllee: Reworded to "To distinguish genuine Wiktionary attrition from a partial scrape" for clarity.

data/scrape/lib/verify_scrape.py
jacksonllee: Left `_pairs` returning a list here: `old_pairs` gets iterated twice -- `_word_to_prons` and `_ordered_unique_words` -- so a lazy generator would be exhausted on the second pass and would need a `list()` at the call site anyway. The two readers feeding it (`_read_working`/`_read_git`) are generators now, so the streaming happens upstream.

data/scrape/lib/verify_scrape.py
jacksonllee: Switched to `operator.itemgetter(2)`.

data/scrape/lib/verify_scrape.py
jacksonllee: Kept `partial_suspected = bool(big)`: it's storing a flag, not gating a branch -- the `--probe` path below can flip it back to `True` -- so it needs to hold the value rather than sit in an `if` block.

tests/test_wikipron/test_scrape.py
jacksonllee: Type annotations added.

kylebgorman (approved):
LGTM

